PR #1, opened 2023-11-01
a code for a class density estimator

JAMSADIQ:
I started a class for density estimator and this is an initial start of this code development.


JAMSADIQ:
request for review

tdent:
I do not see why you have put the module code file into the 'bin' folder or why you have put __init__.py files into the 'bin' and 'examples' folders.  We want the 'bin' folder to hold executables, not classes or modules, so we will not import anything from there.  We want the 'examples' folder to hold scripts, data etc. for running on test examples. 
The 'popde' folder is for the modules and classes, as in the example here https://docs.python-guide.org/writing/structure/#sample-repository

JAMSADIQ:
I removed the extra init files. work in progress for more changes

popde/density_estimate.py
```diff
@@ -0,0 +1,92 @@
+# First script for iterative KDE
+from awkde import GaussianKDE  #we will need our own code 
+
+
+class  densityestimator(object):
```
tdent, line 5: See https://realpython.com/python-pep8/#naming-styles 
The name of the class should use CamelCase
Also, if this is supposed to be a class that does _adaptive_ KDE with cross validation optimization, the name should reflect that
As I wrote in the development plan, we should have a much more basic density estimator class which _just_ calculates a KDE without any other functionality, what you have written is not that, so please give this a suitable name

popde/density_estimate.py
```diff
@@ -0,0 +1,92 @@
+# First script for iterative KDE
+from awkde import GaussianKDE  #we will need our own code
```
tdent, line 2: The comment is not correct, as we might use or write new or different code to produce KDEs. 
Also we should not assume that any KDE is going to use awkde. I would prefer to only import awkde inside class methods that use it

popde/density_estimate.py
```diff
@@ -0,0 +1,92 @@
+# First script for iterative KDE
```
tdent, line 1: remove please, comments should not concern what was 'first' or not, or what we might use code for, they should describe what is _in_ the code

tdent:
For the density estimate module to be able to be installed you do need to add an `__init__.py` file. Please read the online documentation about where and how to do this

popde/density_estimate.py
```diff
@@ -0,0 +1,91 @@
+from awkde import GaussianKDE 
+
+
+class  DensityEstimator(object):
```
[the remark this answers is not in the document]
tdent: In particular, the development plan says that we should also have a class for a general, simple density estimator which is _not_ adaptive or optimized etc.
So that will be another class in addition to the one you wrote here, and the name DensityEstimator might be suitable for that general, simple class, but it is not suitable for the one you have written.

popde/density_estimate.py
tdent: You don't seem to have read my request not to import awkde at the start of the module (as not all of the classes will need to use it) but only inside classes/methods that need it.

JAMSADIQ: I have removed the comment and at the moment import awkde inside class method. Once I have KDE coded this import will be replace or I will modify this method in the class.

popde/density_estimate.py

popde/density_estimate.py
```diff
@@ -0,0 +1,88 @@
+class  SimpleGaussianKDE(object):
```
tdent, line 1: I still do not understand why you are giving the class this name.

tdent:
@JAMSADIQ ok, now that you have responded to all the comments I will also edit the file to clean up the formatting and wording a bit, then it will be ready to merge

tdent:
I am deleting 'code.py' in the bin/ folder because it is not a useful name for a script or executable - you would never call a specific script 'code.py', no-one would know what it does.

tdent:
I made quite a lot of changes, can you have a look?

tdent:
Just to see if you understand what has changed

PR #2, opened 2023-11-20
development for simple Gaussian KDE

JAMSADIQ:
These changes will be for simple Gaussian KDE. At the moment I just add a simple class with two methods to prepare and evaluate KDE given an array for training and bandwidth choice for Gaussian Kernels.


tdent:
It is not informative to call this PR 'Issue2', that does not tell someone reading it what is the content of the change. 
Also can you fix the typo in the name (gaussain)

PR #3, opened 2023-11-20
adding a new class for simple gaussian kde

JAMSADIQ:
I have added a new class which can produce a simple Gaussian KDE.

### DIFF
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -52,7 +52,7 @@ class SimpleKernelDensityEstimation:
 
 
    def check_dimensionality(self):
-        if len(self.data.shape) != len(self.dim_names):
+        if len(self.data.shape[0]) != len(self.dim_names):
             raise ValueError("Dimensionality of data array does not match the number of dimension names.")
 
 
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -26,21 +26,6 @@ class SimpleKernelDensityEstimation:
     def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
         """
         Initialize the KernelDensityEstimation object.
-
-        Parameters:
-        -----------
-        data : array-like
-            The input data for which KDE will be estimated.
-
-        bandwidth : float, optional (default=1.0)
-            Bandwidth parameter controlling the width of the kernels.
-            Larger bandwidths lead to smoother density estimates.
-
-        kernel : choice of kernel, default='gaussian'
-
-        dim_names : list of the dimension names (m1,z, chi_eff) 
-             If dim_names is not Nonecheck that the dimensionality
-             of the data array matches with len(dim_names)
         """
         self.data = np.asarray(data) 
         self.bandwidth = bandwidth
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -22,7 +22,6 @@ class SimpleKernelDensityEstimation:
     new_data = np.array(np.linspace(-3, 3, 100).tolist(), np.linspace(-3, 3, 100).tolist())
     density_values = kde.evaluate(new_data)
     """
-
     def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
         """
         Initialize the KernelDensityEstimation object.
@@ -34,7 +33,6 @@ class SimpleKernelDensityEstimation:
         self.dim_names = dim_names
         if dim_names is not None:
             self.check_dimensionality()
-
 
    def check_dimensionality(self):
         if len(self.data.shape[0]) != len(self.dim_names):
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -26,8 +26,8 @@ class SimpleKernelDensityEstimation:
         """
         Initialize the KernelDensityEstimation object.
         data: array-like, shape (n_samples, n_features)
-              points of the data define each kernel position
-              each row is a point, eachcolumn is a feature.
+               points of the data define each kernel position
+               each row is a point, eachcolumn is a feature.
         """
         if len(data.shape) != 2:
             raise ValueError("data must have shape (n_samples, n_features).")
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 class SimpleKernelDensityEstimation:
     """
     Fit and evaluate general Kernel Density Estimation (KDE) 
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -22,7 +22,7 @@ class SimpleKernelDensityEstimation:
     density_values = kde.evaluate(new_data)
     """
 
-    def __init__(self, data, bandwidth=1.0, kernel='gaussian'):
+    def __init__(self, data, bandwidth=1.0, kernel=['gaussian']):
         """
         Initialize the KernelDensityEstimation object.
 
@@ -42,10 +42,10 @@ class SimpleKernelDensityEstimation:
 
         self.data = np.asarray(data)
         self.bandwidth = bandwidth
-        if self.kernel == 'gaussian':
+        if self.kernel in kernel:
             self.kernel = kernel
         else:
-            raise ValueError("Unsupported kernel. Supported kernels: 'gaussian',")
+            raise NotImplementedError("Unsupported kernel. Supported kernels: 'gaussian',")
 
     def evaluate(self, points):
         """
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -12,7 +12,7 @@ class SimpleKernelDensityEstimation:
     evaluate(points):
         Evaluate the KDE at given data points.
     """
-    def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
+    def __init__(self, data, bandwidth=1.0, dim_names=None):
         """
         Initialize the KernelDensityEstimation object.
         data: array-like, shape (n_samples, n_features)
@@ -27,9 +27,8 @@ class SimpleKernelDensityEstimation:
 
         self.data = np.asarray(data) 
         self.bandwidth = bandwidth
-        self.kernel = kernel 
-
         self.dim_names = dim_names
+
         if dim_names is not None:
             self.check_dimensionality()
 
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -15,11 +15,14 @@ class SimpleKernelDensityEstimation:
     Examples:
     ---------
     # Create a KDE instance and fit it to data
-    data = np.random.randn((100, 100))
-    kde = SimpleGaussianKernelDensityEstimation(data)
+    data = np.random.randn(100, 100)
+    kde = SimpleKernelDensityEstimation(data)
 
     # Evaluate the KDE at new data points
-    new_data = np.array(np.linspace(-3, 3, 100).tolist(), np.linspace(-3, 3, 100).tolist())
+    x = np.linspace(0., 1., 100)
+    y = np.linspace(0., 1., 100)
+    XX, YY = np.meshgrid(x, y)
+    new_data = np.array(list(map(np.ravel, [XX, YY]))).T
     density_values = kde.evaluate(new_data)
     """
     def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
@@ -43,11 +46,11 @@ class SimpleKernelDensityEstimation:
         if dim_names is not None:
             self.check_dimensionality()
 
-   def check_dimensionality(self):
-       """
-       Check if the dimension of KDE matches the parameters on which we evaluate KDE.
+    def check_dimensionality(self):
+        """
+        Check if the dimension of KDE matches the parameters on which we evaluate KDE.
         Raises ValueError
-       """
+        """
         if len(self.data.shape[0]) != len(self.dim_names):
             raise ValueError("Dimensionality of data array does not match the number of dimension names.")
 
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -34,8 +34,7 @@ class SimpleKernelDensityEstimation:
 
     def check_dimensionality(self):
         """
-        Check if the dimension of KDE matches the parameters on which we evaluate KDE.
-        Raises ValueError
+        Check if the dimension of training data matches the number of parameter names.
         """
         if len(self.data.shape[0]) != len(self.dim_names):
             raise ValueError("Dimensionality of data array does not match the number of dimension names.")
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -1,28 +1,6 @@
-class SimpleGaussianKernelDensityEstimation:
+class SimpleKernelDensityEstimation:
     """
-    Kernel Density Estimation (KDE) using Gaussian Kernels
-
-    Parameters:
-    -----------
-    data : array-like
-        The input data for which KDE will be estimated.
-
-    bandwidth : float, optional (default=1.0)
-        Bandwidth parameter controlling the width of the kernels.
-
-    kernel : str, optional (default='gaussian')
-        The kernel function used for smoothing. Supported kernels: 'gaussian',
-
-    Attributes:
-    -----------
-    data : array-like
-        The input data for which KDE is estimated.
-
-    bandwidth : float
-        Bandwidth parameter controlling the width of the kernels.
-
-    kernel : str
-        The kernel function used for smoothing.
+    Kernel Density Estimation (KDE) using Kernels
 
     Methods:
     --------
@@ -66,28 +44,6 @@ class SimpleGaussianKernelDensityEstimation:
         self.bandwidth = bandwidth
         self.kernel = kernel
 
-    def fit(self, data=None, bandwidth=None, kernel=None):
-        """
-        Fit the KDE to new data.
-
-        Parameters:
-        -----------
-        data : array-like, optional
-            The input data for which KDE will be estimated. If None, the data provided during initialization is used.
-
-        bandwidth : float, optional
-            Bandwidth parameter controlling the width of the kernels.
-            Larger bandwidths lead to smoother density estimates. If None, the bandwidth provided during initialization is used.
-
-        kernel : str, optional
-            The kernel function used for smoothing. If None, the kernel provided during initialization is used.
-        """
-        if data is not None:
-            self.data = np.asarray(data)
-        if bandwidth is not None:
-            self.bandwidth = bandwidth
-        if kernel is not None:
-            self.kernel = kernel
 
     def evaluate(self, points):
         """
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -2,7 +2,7 @@ import numpy as np
 
 class SimpleKernelDensityEstimation:
     """
-    Fit and evaluate general Kernel Density Estimation (KDE) 
+    Fit and evaluate multi-dimensional Kernel Density Estimation (KDE) 
 
     Methods:
     --------
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -25,7 +25,13 @@ class SimpleKernelDensityEstimation:
     def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
         """
         Initialize the KernelDensityEstimation object.
+        data: array-like, shape (n_samples, n_features)
+              points of the data define each kernel position
+              each row is a point, eachcolumn is a feature.
         """
+        if len(data.shape) != 2:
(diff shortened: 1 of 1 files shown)


popde/density_estimate.py
```diff
@@ -1,3 +1,128 @@
+class SimpleGaussianKernelDensityEstimation:
+    """
+    Kernel Density Estimation (KDE) using Gaussian Kernels
+
+    Parameters:
```
tdent, line 5: If the parameters are given in the init method, they should not be described here as well. There is no point having the same comments twice

popde/density_estimate.py
```diff
@@ -1,3 +1,128 @@
...
+    data : array-like
+        The input data for which KDE will be estimated.
+
+    bandwidth : float, optional (default=1.0)
+        Bandwidth parameter controlling the width of the kernels.
+
+    kernel : str, optional (default='gaussian')
+        The kernel function used for smoothing. Supported kernels: 'gaussian',
+
+    Attributes:
```
tdent, line 16: I don't know if it is worth writing all of the attributes in a class docstring. Is this recommended anywhere?

popde/density_estimate.py
```diff
@@ -1,3 +1,128 @@
+class SimpleGaussianKernelDensityEstimation:
```
tdent, line 1: I suggest just SimpleKDE, since the kernel is allowed to be something that is not 'gaussian', which would be confusing if 'Gaussian' is in the name

popde/density_estimate.py
```diff
@@ -1,3 +1,128 @@
...
+    """
+    Kernel Density Estimation (KDE) using Gaussian Kernels
```
tdent, line 3: again, if the kernel is not necessarily Gaussian it should not be stated here that it is

popde/density_estimate.py
```diff
@@ -1,3 +1,128 @@
...
+        kernel : str, optional (default='gaussian')
+            The kernel function used for smoothing.
+        """
+        import numpy as np
+
+        self.data = np.asarray(data)
+        self.bandwidth = bandwidth
+        self.kernel = kernel
+
+    def fit(self, data=None, bandwidth=None, kernel=None):
```
tdent, line 69: How is this different from the `__init__` method?  They both set the values of the data, bandwidth and kernel attributes.

popde/density_estimate.py
```diff
@@ -1,3 +1,128 @@
...
+            The estimated density values at the given data points.
+        """
+        import scipy
+        from scipy.stats import norm
+
+        points = np.asarray(points)
+        if self.kernel == 'gaussian':
+            kernel_function = norm(loc=0, scale=self.bandwidth).pdf
+        else:
+            raise ValueError("Unsupported kernel. Supported kernels: 'gaussian',")
```
tdent, line 113: I think `NotImplementedError` would be better .. also, it would be good if the class has an attribute `self.kernels` that is initialized with a list which can just be `['gaussian']` at present

tdent: then you can use this list to make the error message

popde/density_estimate.py
```diff
@@ -1,3 +1,128 @@
...
+        points : array-like
+            The data points at which the KDE will be evaluated.
+
+        Returns:
+        --------
+        density_values : array-like
+            The estimated density values at the given data points.
+        """
+        import scipy
+        from scipy.stats import norm
```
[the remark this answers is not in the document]
tdent: Or maybe this class should not actually have functions that do evaluation etc., it should be a 'empty class' that just has the data structure and names of standard methods but the evaluation is not implemented .. then actual implementations of KDE (like scipy, kdepy, awkde) can be _subclasses_ of this one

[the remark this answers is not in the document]
tdent: I guess we can leave this as it is but note that this implementation is probably going to be very inefficient and we will want to use a different one for actual analysis

popde/density_estimate.py
```diff
@@ -1,3 +1,128 @@
...
+        points = np.asarray(points)
+        if self.kernel == 'gaussian':
+            kernel_function = norm(loc=0, scale=self.bandwidth).pdf
+        else:
+            raise ValueError("Unsupported kernel. Supported kernels: 'gaussian',")
+
+        density_values = np.zeros_like(points, dtype=float)
+
+        for data_point in self.data:
+            density_values += kernel_function((points - data_point) / self.bandwidth)
```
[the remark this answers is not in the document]
tdent: also, is it better to

tdent:
in general, the comments here seem to be too repetitive - there is no point typing the same thing twice for the same attribute of the same class.  But if the 'fit' method can be eliminated we save a lot of repetition

popde/density_estimate.py
JAMSADIQ: Yes, this was repetitive and removed now.

popde/density_estimate.py

popde/density_estimate.py
JAMSADIQ: renamed

popde/density_estimate.py
JAMSADIQ: fix this comment

popde/density_estimate.py
JAMSADIQ: fit is removed now.

popde/density_estimate.py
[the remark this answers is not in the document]
JAMSADIQ: I will try to use scikitlearn  for simple kde., work will need to be done for this

popde/density_estimate.py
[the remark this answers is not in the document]
JAMSADIQ: One dimensional at the moment

popde/density_estimate.py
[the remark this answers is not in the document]
tdent: well, I don't mind if you either keep the function you wrote which only does a 1d KDE, or just make it an empty 'shell' which gives a NotImplementedError (eg here https://github.com/gwastro/pycbc/blob/master/pycbc/events/stat.py#L106) and then have subclasses for various different implementations. 
I don't think we should necessarily always have scikit-learn for the 'simple' class.

tdent:
If you think you have addressed a comment, can you press the 'Resolve conversation' button?

popde/density_estimate.py
JAMSADIQ: Implemented and seems to me its working.

popde/density_estimate.py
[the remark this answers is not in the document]
JAMSADIQ: ok

popde/density_estimate.py
```diff
@@ -1,3 +1,83 @@
+class SimpleKernelDensityEstimation:
+    """
+    Kernel Density Estimation (KDE) using Kernels
```
tdent, line 3: no point to say that KDE uses kernels I think .. 
```suggestion
    Fit and evaluate general Kernel Density Estimation (KDE)
```

popde/density_estimate.py
```diff
@@ -1,3 +1,83 @@
...
+    density_values = kde.evaluate(new_data)
+    """
+
+    def __init__(self, data, bandwidth=1.0, kernel=['gaussian']):
+        """
+        Initialize the KernelDensityEstimation object.
+
+        Parameters:
+        -----------
+        data : array-like
```
tdent, line 31: Can this be multi-dimensional?  I think we want all the classes and methods to be multidimensional from the beginning otherwise there is no point writing new code.

For 1d KDE we already have a complete solution in awkde, the point of having this new package is to deal with issues where we have 2 / 3 / 4 different dimensions in population analysis.

So can you build in the use of multi dimensions from the beginning

popde/density_estimate.py
```diff
@@ -1,3 +1,83 @@
...
+    data = np.random.randn(100)
+    kde = SimpleGaussianKernelDensityEstimation(data)
+    kde.fit()
+
+    # Evaluate the KDE at new data points
+    new_data = np.linspace(-3, 3, 100)
+    density_values = kde.evaluate(new_data)
+    """
+
+    def __init__(self, data, bandwidth=1.0, kernel=['gaussian']):
```
tdent, line 25: Also I'd suggest to allow an optional kwarg which is a list of the dimension names (ie names of variables like mass, redshift, chi_eff) and then if it is given to check that the dimensionality of the data array is equal to the number of dimension names.

popde/density_estimate.py

popde/density_estimate.py
JAMSADIQ: added a bit of  code. need to check if this make sense to you?

popde/density_estimate.py
JAMSADIQ: ok

popde/density_estimate.py
```diff
@@ -1,3 +1,97 @@
...
+            Larger bandwidths lead to smoother density estimates.
+
+        kernel : list of str, optional (default=['gaussian'])
+            The kernel function used for smoothing.
+
+        dim_names : list of the dimension names (m1,z, chi_eff) 
+             If dim_names is not Nonecheck that the dimensionality
+             of the data array matches with len(dim_names)
+        """
+        import numpy as np
```
tdent, line 45: I think we will always use numpy, so please import this at the start of the file

popde/density_estimate.py
```diff
@@ -1,3 +1,97 @@
...
+        else:
+            raise NotImplementedError("Unsupported kernel. Supported kernels: 'gaussian',")
+
+        self.dim_names = dim_names
+        if dim_names is not None:
+            self.check_dimensionality()
+
+
+   def check_dimensionality(self):
+        if len(self.data.shape) != len(self.dim_names):
```
tdent, line 60: have you checked this?  ie what is `len(self.data.shape)` ?
For example if you are using 3d data, what should `self.data.shape` be ?

popde/density_estimate.py
```diff
@@ -1,3 +1,97 @@
...
+    fit(data, bandwidth=None, kernel=None):
+        Fit the KDE to new data.
+
+    evaluate(points):
+        Evaluate the KDE at given data points.
+
+    Examples:
+    ---------
+    # Create a KDE instance and fit it to data
+    data = np.random.randn(100)
```
tdent, line 16: this example is only 1d data.  I think we should have an example that uses multi-dimensional data, eg 2d
Also, the example code does `kde.fit()` which doesn't exist.

popde/density_estimate.py
```diff
@@ -1,3 +1,97 @@
...
+        dim_names : list of the dimension names (m1,z, chi_eff) 
+             If dim_names is not Nonecheck that the dimensionality
+             of the data array matches with len(dim_names)
+        """
+        import numpy as np
+
+        self.data = np.asarray(data)
+        self.bandwidth = bandwidth
+        if self.kernel in kernel:
+            self.kernel = kernel[0]#may need to fix this
```
tdent, line 50: this cannot work because `kernel` is not defined. You need to define some variable, attribute etc which would be a list of available kernels

popde/density_estimate.py
JAMSADIQ: fixed this

popde/density_estimate.py
JAMSADIQ: I checked and if I have a data in shape like (10, 10, 10) , the len will be 3. I am not sure if we can have data shape (10, 10, 10, 1)

popde/density_estimate.py
JAMSADIQ: fixed this

popde/density_estimate.py
JAMSADIQ: fixed it and seems to work on an example

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
+import numpy as np
+
+class SimpleKernelDensityEstimation:
+    """
+    Fit and evaluate general Kernel Density Estimation (KDE)
```
tdent, line 5:
```suggestion
    Fit and evaluate multi-dimensional Kernel Density Estimation (KDE)
```

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
...
+             of the data array matches with len(dim_names)
+        """
+        self.data = np.asarray(data) 
+        self.bandwidth = bandwidth
+        self.kernel = kernel 
+
+        self.dim_names = dim_names
+        if dim_names is not None:
+            self.check_dimensionality()
+
```
tdent, line 52: Only 1 space between methods
```suggestion
```

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
...
+
+   def check_dimensionality(self):
+        if len(self.data.shape[0]) != len(self.dim_names):
+            raise ValueError("Dimensionality of data array does not match the number of dimension names.")
+
```
tdent, line 57:
```suggestion

```

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
+import numpy as np
+
+class SimpleKernelDensityEstimation:
+    """
+    Fit and evaluate general Kernel Density Estimation (KDE) 
+
+    Methods:
+    --------
+    fit(data, bandwidth=None, kernel=None):
```
tdent, line 9: The fit method does not exist any more. However you added check dimensionality

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
...
+
+        bandwidth : float, optional (default=1.0)
+            Bandwidth parameter controlling the width of the kernels.
+            Larger bandwidths lead to smoother density estimates.
+
+        kernel : choice of kernel, default='gaussian'
+
+        dim_names : list of the dimension names (m1,z, chi_eff) 
+             If dim_names is not Nonecheck that the dimensionality
+             of the data array matches with len(dim_names)
```
tdent, line 43:
```suggestion
        dim_names : sequence of dimension names, e.g. ('m1', 'z', 'chi_eff') 
```
You should just document the kwarg here, put a comment at the dimensionality check method instead

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
...
+        self.bandwidth = bandwidth
+        self.kernel = kernel 
+
+        self.dim_names = dim_names
+        if dim_names is not None:
+            self.check_dimensionality()
+
+
+   def check_dimensionality(self):
+        if len(self.data.shape[0]) != len(self.dim_names):
```
tdent, line 55: the choice of shape[0] rather than shape[1] implies that the shape of the data has to be (ndim, nevents) not (nevents, ndim)? 
Are we sure about this?

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
...
+    density_values = kde.evaluate(new_data)
+    """
+
+    def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
+        """
+        Initialize the KernelDensityEstimation object.
+
+        Parameters:
+        -----------
+        data : array-like
```
tdent, line 32: We should make a decision and document what the shape should be, either (nevents, ndim) or (ndim, nevents)

popde/density_estimate.py
tdent: (also, the names have to be strings!

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
...
+        Returns:
+        --------
+        density_values : array-like
+            The estimated density values at the given data points.
+        """
+        import scipy
+        from scipy.stats import norm
+
+        points = np.asarray(points)
+        kernel_function = norm(loc=0, scale=self.bandwidth).pdf
```
tdent, line 77: does this work for multidimensional data or only 1d?

popde/density_estimate.py

popde/density_estimate.py
JAMSADIQ: fixed

popde/density_estimate.py
```diff
@@ -1,3 +1,92 @@
...
+        kernel : choice of kernel, default='gaussian'
+
+        dim_names : list of the dimension names (m1,z, chi_eff) 
+             If dim_names is not Nonecheck that the dimensionality
+             of the data array matches with len(dim_names)
+        """
+        self.data = np.asarray(data) 
+        self.bandwidth = bandwidth
+        self.kernel = kernel 
+
```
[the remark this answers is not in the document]
JAMSADIQ: need to fix this after changing the code

popde/density_estimate.py
JAMSADIQ: this is fixed

popde/density_estimate.py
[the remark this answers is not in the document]
JAMSADIQ: ok

popde/density_estimate.py
```diff
@@ -1,3 +1,83 @@
...
+    # Evaluate the KDE at new data points
+    new_data = np.array(np.linspace(-3, 3, 100).tolist(), np.linspace(-3, 3, 100).tolist())
+    density_values = kde.evaluate(new_data)
+    """
+    def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
+        """
+        Initialize the KernelDensityEstimation object.
+        data: array-like, shape (n_samples, n_features)
+               points of the data define each kernel position
+               each row is a point, eachcolumn is a feature.
```
tdent, line 30:
```suggestion
               each row is a point, each column is a parameter.
```

popde/density_estimate.py
```diff
@@ -1,3 +1,83 @@
...
+            The data points at which the KDE will be evaluated.
+
+        Returns:
+        --------
+        density_values : array-like
+            The estimated density values at the given data points.
+            using standard gaussian distribution
+        """
+        import scipy
+        from scipy.stats import gaussian_kde
```
tdent, line 71: do you need both of these imports?

popde/density_estimate.py
```diff
@@ -1,3 +1,83 @@
...
+        --------
+        density_values : array-like
+            The estimated density values at the given data points.
+            using standard gaussian distribution
+        """
+        import scipy
+        from scipy.stats import gaussian_kde 
+
+        points = np.asarray(points).T
+        kernel_function = gaussian_kde(self.data.T, bw_method=self.bandwidth)
```
tdent, line 74: So this initializes a scipy object which can be used to evaluate the KDE later if it is set as an attribute of the SimpleKernelDensityEstimation object, ie we could make the kernel function once and effectively save it. 
I guess we can consider this in a later improvement

popde/density_estimate.py
```diff
@@ -1,3 +1,82 @@
...
+    check_dimensionality():
+        check if the data matches the dimensions of KDE.
+
+    evaluate(points):
+        Evaluate the KDE at given data points.
+
+    Examples:
+    ---------
+    # Create a KDE instance and fit it to data
+    data = np.random.randn((100, 100))
```
tdent, line 18: is this correct?   Have you tried running this example?  How many dimensions do you want to make the data with?

JAMSADIQ: resolved

popde/density_estimate.py
JAMSADIQ: fixed

popde/density_estimate.py
```diff
@@ -1,3 +1,71 @@
...
+
+    Methods:
+    --------
+    check_dimensionality():
+        check if the data matches the dimensions of KDE.
+
+    evaluate(points):
+        Evaluate the KDE at given data points.
+    """
+    def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
```
tdent, line 15: I am not sure why you have the `kernel` kwarg, it is not documented and does not do anything. Probably best to remove it

popde/density_estimate.py
```diff
@@ -1,3 +1,71 @@
...
+        """
+        Initialize the KernelDensityEstimation object.
+        data: array-like, shape (n_samples, n_features)
+               points of the data define each kernel position
+               each row is a point, each column is a parameter.
+        kwargs:
+            dim_names : sequence of dimension names, e.g. ('m1', 'z', 'chi_eff')
```
tdent, line 22: add a line documenting bandwidth please

popde/density_estimate.py
```diff
@@ -1,3 +1,71 @@
...
+        self.kernel = kernel 
+
+        self.dim_names = dim_names
+        if dim_names is not None:
+            self.check_dimensionality()
+
+    def check_dimensionality(self):
+        """
+        Check if the dimension of KDE matches the parameters on which we evaluate KDE.
+        Raises ValueError
```
tdent, line 39: It is obvious from the code that it raises a ValueError, no need for an extra comment line
```suggestion
        Check if the dimension of training data matches the number of parameter names.
```

popde/density_estimate.py
JAMSADIQ: added

popde/density_estimate.py

tdent (approved):

PR #6, opened 2023-12-09
Allow different backends, add kdepy class

tdent:
Separate out a fit method, call it at initialization and allow for different backends via an init kwarg, also allow for different evaluate methods

### DIFF
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -2,7 +2,7 @@ import numpy as np
 
 class SimpleKernelDensityEstimation:
     """
-    Fit and evaluate multi-dimensional Kernel Density Estimation (KDE) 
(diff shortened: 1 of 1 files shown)


popde/density_estimate.py
```diff
@@ -12,37 +12,62 @@ class SimpleKernelDensityEstimation:
     evaluate(points):
         Evaluate the KDE at given data points.
     """
-    def __init__(self, data, bandwidth=1.0, dim_names=None):
+    def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
```
JAMSADIQ, line 15: ok

JAMSADIQ (approved):
ok.

Issue #4, opened 2023-12-08
Add an example in documentation 


JAMSADIQ:
Add an example for the use of class SimpleKernelDensityEstimation after verification

tdent:
If this has been addressed, please close the issue

PR #10, opened 2023-12-12
Documentation of an example for SimpleKernelDensityEstimation

JAMSADIQ:
To complete issue #4 of adding an example in documentation.

### DIFF
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -52,7 +52,7 @@ class SimpleKernelDensityEstimation:
 
 
    def check_dimensionality(self):
-        if len(self.data.shape) != len(self.dim_names):
+        if len(self.data.shape[0]) != len(self.dim_names):
             raise ValueError("Dimensionality of data array does not match the number of dimension names.")
 
 
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -26,21 +26,6 @@ class SimpleKernelDensityEstimation:
     def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
         """
         Initialize the KernelDensityEstimation object.
-
-        Parameters:
-        -----------
-        data : array-like
-            The input data for which KDE will be estimated.
-
-        bandwidth : float, optional (default=1.0)
-            Bandwidth parameter controlling the width of the kernels.
-            Larger bandwidths lead to smoother density estimates.
-
-        kernel : choice of kernel, default='gaussian'
-
-        dim_names : list of the dimension names (m1,z, chi_eff) 
-             If dim_names is not Nonecheck that the dimensionality
-             of the data array matches with len(dim_names)
         """
         self.data = np.asarray(data) 
         self.bandwidth = bandwidth
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -22,7 +22,6 @@ class SimpleKernelDensityEstimation:
     new_data = np.array(np.linspace(-3, 3, 100).tolist(), np.linspace(-3, 3, 100).tolist())
     density_values = kde.evaluate(new_data)
     """
-
     def __init__(self, data, bandwidth=1.0, kernel='gaussian', dim_names=None):
         """
         Initialize the KernelDensityEstimation object.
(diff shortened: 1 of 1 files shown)


popde/density_estimate.py
```diff
@@ -18,9 +18,39 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
                points of the data define each kernel position
                each row is a point, each column is a parameter.
         kwargs:
-            bandwidth : The bandwidth of the kernel used for smoothing
+           bandwidth : The bandwidth of the kernel used for smoothing
             dim_names : sequence of dimension names, e.g. ('m1', 'z', 'chi_eff')
                         values must be strings
+        Example:
+        >>> #For two dimensional case
```
tdent, line 25: Can you remove the `>>>` please, someone who wants to run the example cannot use them as input so they would have to remove them anyway

popde/density_estimate.py
```diff
@@ -18,9 +18,39 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
...
+        >>> XX, YY = np.meshgrid(x, y)
+        >>> eval_pts = np.array(list(map(np.ravel, [XX, YY]))).T
+        >>> dx2, dy2 = (x[1] - x[0]) / 2., (y[1] - y[0]) / 2.
+        >>> bx = np.concatenate((x - dx2, [x[-1] + dx2]))
+        >>> by = np.concatenate((y - dy2, [y[-1] + dy2]))
+        >>> zz = kde.evaluate(grid_pts)
+        >>> ZZ = zz.reshape(XX.shape)
+        >>> import matplotlib.pyplot as plt
+        >>> from matplotlib.colors import LogNorm
+        >>> plt.figure()
```
tdent, line 47: I don't think this line is necessary

popde/density_estimate.py
```diff
@@ -18,9 +18,39 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
...
             dim_names : sequence of dimension names, e.g. ('m1', 'z', 'chi_eff')
                         values must be strings
+        Example:
+        >>> #For two dimensional case
+        >>> rndgen = np.random.RandomState(seed=3575)
+        >>> mean1, sigma1 = 3.0, .25
+        >>> mean2, sigma2 = 14.0, 1.5
+        >>> n_samples = 1000
+        >>> sample1 = rndgen.normal(mean1, sigma1, size=n_samples)
+        >>> sample2 = rndgen.normal(mean2, sigma2, size=n_samples
```
tdent, line 31: This line seems to have a typo

popde/density_estimate.py
```diff
@@ -18,9 +18,39 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
...
+        >>> sample = np.vstack((sample1, sample2)).T #shape of data (n_points, n_features)
```
tdent, line 32: if you have to transpose the result of vstack, then vstack is not the correct function. Look for a np array function that gives the correct shape directly

popde/density_estimate.py
```diff
@@ -18,9 +18,39 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
...
+        >>> kde = SimpleKernelDensityEstimation(sample, dim_names=['mass1', 'mass2'])
+        >>> minx, maxx = np.amin(sample[:, 0]), np.amax(sample[:, 0])
+        >>> miny, maxy = np.amin(sample[:, 1]), np.amax(sample[:, 1])
+        >>> x = np.linspace(minx, maxx, 100)
+        >>> y = np.linspace(miny, maxy, 100)
+        >>> XX, YY = np.meshgrid(x, y)
+        >>> eval_pts = np.array(list(map(np.ravel, [XX, YY]))).T
```
tdent, line 39: This line is much too complicated to understand, please write it in a simpler way. Is it necessary to use `list(map(..))` ?  If the result of np.ravel on XX and YY is a numpy array, why would you need to make a list of the two and then convert it to an array and then take the transpose to get what you want?  Can you look for a numpy function that just converts ravel(XX) and ravel(YY) into the format you need?

popde/density_estimate.py
```diff
@@ -18,9 +18,39 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
...
+        >>> dx2, dy2 = (x[1] - x[0]) / 2., (y[1] - y[0]) / 2.
+        >>> bx = np.concatenate((x - dx2, [x[-1] + dx2]))
```
tdent, line 41: I cannot see what you are doing here.  Either make a comment to explain or simplify the code so it is clearer.

PR #11, opened 2023-12-13
Example doc

JAMSADIQ:
fixed comments as suggested for issue #4


popde/density_estimate.py
```diff
@@ -18,9 +18,37 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
...
+          sample1 = rndgen.normal(mean1, sigma1, size=n_samples)
+          sample2 = rndgen.normal(mean2, sigma2, size=n_samples)
+          sample = np.column_stack((sample1, sample2)) #shape of data (n_points, n_features)
+          kde = SimpleKernelDensityEstimation(sample, dim_names=['mass1', 'mass2'])
+          minx, maxx = np.amin(sample[:, 0]), np.amax(sample[:, 0])
+          miny, maxy = np.amin(sample[:, 1]), np.amax(sample[:, 1])
+          x = np.linspace(minx, maxx, 100)
+          y = np.linspace(miny, maxy, 100)
+          XX, YY = np.meshgrid(x, y)
+          eval_pts = np.column_stack((XX.reshape(-1), YY.reshape(-1)))
```
tdent, line 41: this is clearer, though I am not immediately sure what reshape(-1) does. If XX.flatten() or XX.ravel() work I would prefer one of these

popde/density_estimate.py
```diff
@@ -18,9 +18,37 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
...
+          #For two dimensional case
+          rndgen = np.random.RandomState(seed=3575)
+          mean1, sigma1 = 3.0, .25
+          mean2, sigma2 = 14.0, 1.5
+          n_samples = 1000
+          sample1 = rndgen.normal(mean1, sigma1, size=n_samples)
+          sample2 = rndgen.normal(mean2, sigma2, size=n_samples)
+          sample = np.column_stack((sample1, sample2)) #shape of data (n_points, n_features)
+          kde = SimpleKernelDensityEstimation(sample, dim_names=['mass1', 'mass2'])
+          minx, maxx = np.amin(sample[:, 0]), np.amax(sample[:, 0])
```
tdent, line 36: isn't it simpler to use min/max of sample1 and sample2 ?

popde/density_estimate.py

popde/density_estimate.py
JAMSADIQ: yes and its changed now.

tdent (approved):

PR #12, opened 2023-12-13
started a module for plotting

JAMSADIQ:
A  script is initiated for plotting of KDE using KDE classes. The goal is to accomplish task described in issue #9.


popde/density_estimate.py
```diff
@@ -207,3 +249,5 @@ def estimation(self, samples, x_eval, alphagrid, bwgrid):
         kdeval = train_eval_kde(samples, x_eval, self.optbw, self.optalpha)
 
         return kdeval, self.optbw, self.optalpha
+
+
```
tdent, line 253:
```suggestion

```

popde/density_estimate.py
```diff
@@ -92,6 +92,48 @@ def evaluate_scipy(self, points):
...
+        points = np.vstack([grid[i].ravel() for i in range(len(grid))]).T
+        density_values = self.evaluate(points).reshape(grid[0].shape)
+
+        if use_log:
+            density_values = density_values*np.exp(grid)#need to fix this
+
+        plot_function(grid, density_values, labels=labels, **plot_kwargs)
+        return 0
+
+
```
tdent, line 136:
```suggestion
```

popde/utils_plot.py
```diff
@@ -0,0 +1,164 @@
...
+sample1 = rndgen.normal(mean1, sigma1, size=n_samples)
+sample2 = rndgen.normal(mean2, sigma2, size=n_samples)
+data= np.column_stack((sample1, sample2)) # shape is (n_points, n_features)
+kde_instance = SimpleKernelDensityEstimation(data, dim_names=['m1', 'z'])
+kde_plotter = KDEPlotter(kde_instance)
+kde_plotter.plot_1d_projection(0)
+kde_plotter.plot_2d_slice(0, 1)
+
+
+class DataVisualizer:
```
tdent, line 111: I don't know what this class should do. The point of creating a class is to organize some information that belongs together and should persist through several operations.

popde/utils_plot.py
```diff
@@ -0,0 +1,164 @@
...
+    cset = ax.contour(xx, yy, kdevals, colors='k')
+    # Label plot
+    ax.clabel(cset, inline=1, fontsize=10)
+    ax.set_xlabel('Y1')
+    ax.set_ylabel('Y0')
+
+    plt.show()
+
+
+class KDEPlotter:
```
tdent, line 24: Some of the functions here could be useful, but I don't see why this is a class. The point of making a class is to organize information that belongs together and should persist for some time. The density estimate class already does this for the data points and the KDE.

tdent:
I can't really tell which parts of code I am supposed to review in this change. It seems to have lots of unfinished or unused extra parts.  Please change the PR so that it contains only functional code, ie code that works and is necessary to produce some plot.  At the moment I cannot properly review it as the changes seem to be for two or three different ways of making plots which conflict with each other. 

Looking at the 'plot' method in the density estimate module, it is not going to do something useful because it creates a new density estimate instance with new (fake?) data.  I don't see what the point of that would be.  We want the plotting methods of a density estimate object to make plots of the data and KDE that are attributes _of that object_.  Not some random new data. 

Can you try just writing a method which makes a 2d contour plot of the KDE for the simple KDE class.  For a 2d data set it will be straightforward, for more dimensions it should get arguments telling it which dimensions are the plot axes, and the values of the other dimensions to make the 'slicing' over.

popde/utils_plot.py
JAMSADIQ: I removed this.

popde/utils_plot.py
JAMSADIQ: removed

JAMSADIQ:
resolved some commits and add a very simple method in kde class for sliced plot.

popde/utils_plot.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+
+def simple2Dplot(xx, yy, kde):
+    """
+    return kde plot in 2D given output
+    """
+    fig = plt.figure()
+    ax = fig.gca()
+    # Contourf plot
```
tdent, line 12: The comment is not informative for someone who does not already know 'contourf'. Suggest 'Filled contour plot'

popde/utils_plot.py
```diff
@@ -0,0 +1,22 @@
...
+    cfset = ax.contourf(xx, yy, kdevals, cmap='Blues')
+    # Contour plot
+    cset = ax.contour(xx, yy, kdevals, colors='k')
+    # Label plot
+    ax.clabel(cset, inline=1, fontsize=10)
+    ax.set_xlabel('Y1')
```
tdent, line 18: why would you want x and y labels called 'Y1' and 'Y0' ?  The dimensions of the KDE already have names assigned to them, those should be used

popde/density_estimate.py
```diff
@@ -93,6 +93,67 @@ def evaluate_scipy(self, points):
...
+        """
+        # Generate a grid for the contour plot
+        dim1_min, dim1_max = np.min(self.data[dim1]), np.max(self.data[dim1])
+        dim2_min, dim2_max = np.min(self.data[dim2]), np.max(self.data[dim2])
+        dim1_grid = np.linspace(dim1_min, dim1_max, num_points)
+        dim2_grid = np.linspace(dim2_min, dim2_max, num_points)
+        xx, yy = np.meshgrid(dim1_grid, dim2_grid)
+        positions = np.column_stack([xx.ravel(), yy.ravel()])
+
+        # If slicing is specified, insert the slice values into the positions array
```
tdent, line 135: I think we should have a check for dimensionality, ie the number of slice parameters / values should be equal to the number of KDE dimensions minus 2

popde/density_estimate.py
```diff
@@ -93,6 +93,67 @@ def evaluate_scipy(self, points):
...
+        if slice_dims is not None and slice_values is not None:
+            for slice_dim, slice_value in zip(slice_dims, slice_values):
+                slice_idx = list(self.data.columns).index(slice_dim)
+                positions = np.insert(positions, slice_idx, slice_value, axis=0)
+
+        # Evaluate the KDE at the grid points need fix here
```
tdent, line 141: what does the comment 'need fix here' mean

popde/density_estimate.py
```diff
@@ -93,6 +93,67 @@ def evaluate_scipy(self, points):
...
+        z = self.kde(positions)
+
+        # Reshape the results for contour plotting
+        zz = z.reshape(xx.shape)
+
+        # Create the contour plot
```
tdent, line 147: why are you not using the function you wrote in the plot_utils module?  the whole point of having a plotting module is to import functions from it here so that the code here in the main module is more concise

popde/utils_plot.py
tdent: Or just leave it to the plotting method inside density_estimate.py to set the labels ..

popde/density_estimate.py
```diff
@@ -93,6 +93,67 @@ def evaluate_scipy(self, points):
...
+        plt.contour(xx, yy, zz, **kwargs)
+        plt.xlabel(dim1)
+        plt.ylabel(dim2)
+        plt.title(f'2D Contour Plot of KDE for {dim1} and {dim2} (Sliced along {slice_dims})')
+        plt.savefig("KDE"+dim1+dim2+"_contour.png")
```
tdent, line 152: this is a reasonable_default name to save to, but it would be good if the method can override this name (eg if we were making several plots at different slice values). So add a plot file name kwarg

popde/density_estimate.py
```diff
@@ -93,6 +93,67 @@ def evaluate_scipy(self, points):
...
+        plt.show()
```
tdent, line 153: as I said for the plot utils function you should not assume this is being run on a system where you can just `show` a plot, this will not work on a cluster.  Bear in mind that this method will be called inside a script so you can do `return fig` here (after creating a fig object) to allow someone to do `show` inside the script if wanted. 

Also there should be the option of _not_ saving the plot, eg if the method is called with file name = None

popde/density_estimate.py
```diff
@@ -93,6 +93,67 @@ def evaluate_scipy(self, points):
...
+
+            # Generate 3D normal distributed data
+            data = np.random.multivariate_normal(mean, covariance_matrix, num_points)
+            parameter = ['m1', 'm2', 'Mc']
+            # Extract individual variables
+            x, y, z = data[:, 0], data[:, 1], data[:, 2]
+            kde = (data)
+
+            # Plot a 2D contour with a slice along the 'z' dimension
+            kde.plot_2d_contour(parameter[0], parameter[1], slice_dims=[parameter[2]], slice_values=[0], num_points=100, colors='blue', alpha=0.5)
```
tdent, line 124: do the kwargs 'colors' and 'alpha' do anything here?

tdent:
Basically this does not quite make sense yet, as the idea of having a plotting module is to import the functions from that module into the main density estimate module. 
Also I would prefer to name the module `plot_utils` (we now have `transform_utils`), hopefully it is not too much trouble to rename it via `git mv`.

popde/density_estimate.py
JAMSADIQ: added this option

popde/density_estimate.py
```diff
@@ -93,6 +93,67 @@ def evaluate_scipy(self, points):
...
+
+        """
+        # Generate a grid for the contour plot
+        dim1_min, dim1_max = np.min(self.data[dim1]), np.max(self.data[dim1])
```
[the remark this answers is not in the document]
JAMSADIQ: done

popde/density_estimate.py
JAMSADIQ: done

popde/density_estimate.py
JAMSADIQ: done

popde/density_estimate.py
```diff
@@ -93,6 +119,63 @@ def evaluate_scipy(self, points):
...
+            kde.plot_2d_contour(parameter[0], parameter[1], slice_dims=[parameter[2]], slice_values=[0], num_points=100)
+
+        """
+        # Generate a grid for the contour plot
+        xx, yy = utils_plot.get_twoD_grid(self.data[dim1], self.data[dim2], num_points=num_points)
+        positions = np.column_stack([xx.ravel(), yy.ravel()])
+
+        #check if slice_dimensions  == dim(KDE) -  2
+        if len(slice_dims) != self.data.shape[1] - 2 :
+            raise ValueError("kde must be computed on more than 2 parameters for slicing")
```
tdent, line 159: ok, but the error message is not very informative.  I would suggest eg `f"With {self.data.shape[1]} KDE dimensions, must specify {self.data.shape[1] - 2} slicing parameters for the plot"`

Also can you put in a check that the length of slice values is equal to the slice parameters

popde/density_estimate.py
```diff
@@ -93,6 +119,63 @@ def evaluate_scipy(self, points):
...
+
+        # If slicing is specified, insert the slice values into the positions array
+        if slice_dims is not None and slice_values is not None:
```
tdent, line 162: if we already checked that these have the same length, then you only need to check that one of them is not None
(nb, if only one is not None there is a problem!)

popde/density_estimate.py
```diff
@@ -93,6 +119,63 @@ def evaluate_scipy(self, points):
...
+            for slice_dim, slice_value in zip(slice_dims, slice_values):
+                slice_idx = list(self.data.columns).index(slice_dim)
+                positions = np.insert(positions, slice_idx, slice_value, axis=0)
+
+        # Evaluate the KDE at the grid points
+        z = self.kde(positions)
```
tdent, line 168: does this work? what does the `self.kde` method do?

popde/density_estimate.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import utils_plot as utils_plot
```
tdent, line 2: why are you doing this with 'as' ? ... what is wrong with `import utils_plot` ?

popde/density_estimate.py
```diff
@@ -93,6 +119,63 @@ def evaluate_scipy(self, points):
...
+        # If slicing is specified, insert the slice values into the positions array
+        if slice_dims is not None and slice_values is not None:
+            for slice_dim, slice_value in zip(slice_dims, slice_values):
+                slice_idx = list(self.data.columns).index(slice_dim)
+                positions = np.insert(positions, slice_idx, slice_value, axis=0)
+
+        # Evaluate the KDE at the grid points
+        z = self.kde(positions)
+
+        # Reshape the results for contour plotting
```
tdent, line 170:
```suggestion
```
Since the function is called 'reshape' the comment doesn't really add anything. I would move this line to after the 'Create the contour plot' comment

popde/utils_plot.py
```diff
@@ -0,0 +1,39 @@
...
+    # Contour line labels
+    ax.clabel(cset, inline=1, fontsize=10)
+    if xlabel:
+        ax.set_xlabel(xlabel)
+    if ylabel:
+        ax.set_ylabel(ylabel)
+    if title:
+        ax.set_title(title)
+    return fig
+
```
tdent, line 25:
```suggestion


```
Two blank lines between functions (though not between class methods)

popde/utils_plot.py
```diff
@@ -0,0 +1,38 @@
...
+def get_twoD_grid(data_dim1, data_dim2, num_points=100):
```
tdent, line 26: It would be nice to have a 'pad' option which puts a slightly bigger range of the coordinate so that we don't get points exactly on the boundaries.  If we define dim1_range as dim1_max - dim1_min,  eg 'pad=0.01' would take the linspace to be between (dim1_min - 0.01 * dim1_range) and (dim1_min + 0.01 * dim1_range) and the same for dim2. 

Also it would be nice to have dim1 range and dim2 range options so that if wanted the user could specify whatever they wanted.

popde/density_estimate.py
```diff
@@ -93,6 +119,62 @@ def evaluate_scipy(self, points):
...
+            kde.plot_2d_contour(parameter[0], parameter[1], slice_dims=[parameter[2]], slice_values=[0], num_points=100)
+
+        """
+        # Generate a grid for the contour plot
+        xx, yy = utils_plot.get_twoD_grid(self.data[dim1], self.data[dim2], num_points=num_points)
+        positions = np.column_stack([xx.ravel(), yy.ravel()])
+
+        #check if slice_dimensions  == dim(KDE) -  2
+        if len(slice_dims) != self.data.shape[1] - 2 :
+            raise ValueError(f"With {self.data.shape[1]} KDE dimensions, must specify {self.data.shape[1] - 2} slicing parameters for the plot")
```
tdent, line 159: also we want to check if the length of slice vals and slice dims is the same, no?

popde/density_estimate.py
```diff
@@ -93,6 +119,62 @@ def evaluate_scipy(self, points):
...
+        # If slicing is specified, insert the slice values into the positions array
+        if slice_values is not None:
+            for slice_dim, slice_value in zip(slice_dims, slice_values):
+                slice_idx = list(self.data.columns).index(slice_dim)
+                positions = np.insert(positions, slice_idx, slice_value, axis=0)
+
+        # Evaluate the KDE at the grid points
+        z = self.kde(positions)
+
+        zz = z.reshape(xx.shape)
```
tdent, line 170: I suggested to move this line to after the comment 'Create the contour plot'

popde/density_estimate.py

popde/density_estimate.py
JAMSADIQ: done

popde/density_estimate.py
JAMSADIQ: replaced

popde/density_estimate.py
JAMSADIQ: added an raise error

popde/density_estimate.py
JAMSADIQ: done

popde/utils_plot.py
JAMSADIQ: added

popde/utils_plot.py
```diff
@@ -0,0 +1,59 @@
+import numpy as np
+import matplotlib.pyplot as plt
+
```
tdent, line 3:
```suggestion
```

popde/utils_plot.py
```diff
@@ -0,0 +1,58 @@
...
+        dim1_min, dim1_max = dim1_range
+    else:
+        dim1_min, dim1_max = np.min(data_dim1), np.max(data_dim1)
+
+    if dim2_range is not None:
+        dim2_min, dim2_max = dim2_range
+    else:
+        dim2_min, dim2_max = np.min(data_dim2), np.max(data_dim2)
+
+    if pad is not None:
```
tdent, line 44: as notes above, these lines can be removed from inside the 'if' switch since the default setting pad=0 will leave the min/max unchanged.

popde/density_estimate.py
```diff
@@ -167,6 +168,67 @@ def evaluate_scipy(self, points):
...
+            raise ValueError(f"Number of slice dimensions must match number of slice values")
+
+        # If slicing is specified, insert the slice values into the positions array
+        if slice_values is not None:
+            for slice_dim, slice_value in zip(slice_dims, slice_values):
+                slice_idx = self.dim_names.index(slice_dim)
+                positions = np.insert(positions, slice_idx, slice_value, axis=1)
+
+        # Evaluate the KDE at the grid points
+        z = self.evaluate_scipy(positions)
```
tdent, line 221: why have you used evaluate_scipy?  the point of writing the self.evaluate() method is to allow for the backend to be set flexibly when the method is initialized.  Is there any reason why just self.evaluate(positions) might not work?

tdent: (for instance if we use the KDEpy class, which subclasses this one, we do not want the plot to be made by using evaluate_scipy)

popde/utils_plot.py
```diff
@@ -0,0 +1,59 @@
...
+    if ylabel:
+        ax.set_ylabel(ylabel)
+    if title:
+        ax.set_title(title)
+    if show_plot==True:
+        plt.show()
+    return fig
+
+
+def get_twoD_grid(data_dim1, data_dim2, num_points=100, pad=None, dim1_range=None, dim2_range=None):
```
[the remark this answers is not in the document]
JAMSADIQ: done

popde/density_estimate.py
JAMSADIQ: done

popde/density_estimate.py
tdent: why does the comment say 'need fix here'? 
```suggestion
        # Evaluate the KDE at the grid points
```

popde/density_estimate.py
```diff
@@ -167,6 +167,67 @@ def evaluate_scipy(self, points):
...
+            # Plot a 2D contour with a slice along the 'z' dimensios
+            fig = kde.plot_2d_contour(parameter[0],parameter[1], 
+                    slice_dims=[parameter[2]], slice_values=[0], num_points=100)
+        """
+        if dim1 not in self.dim_names or dim2 not in self.dim_names:
+            raise ValueError("Invalid dimension names")
+        #get the samples along the dimensions 
+        idx_dim1 = self.dim_names.index(dim1)
+        idx_dim2 = self.dim_names.index(dim2)
+        # Generate a grid for the contour plot
```
tdent, line 203: Space for readability
```suggestion

        # Generate a grid for the contour plot
```

popde/density_estimate.py
```diff
@@ -167,6 +167,67 @@ def evaluate_scipy(self, points):
...
+        xx, yy = utils_plot.get_twoD_grid(self.data[:, idx_dim1], self.data[:, idx_dim2], num_points=num_points)
+        positions = np.column_stack([xx.ravel(), yy.ravel()])
+        #check if slice_dimensions  == dim(KDE) -  2
```
tdent, line 206: Comment is pointless, as it repeats exactly what is in the next line of code
```suggestion
```

popde/density_estimate.py
```diff
@@ -167,6 +167,67 @@ def evaluate_scipy(self, points):
...
+        if len(slice_dims) != self.data.shape[1] - 2:
```
tdent, line 207: Put this check and the other one before anything else is done, ie just after the dimension names check. If there is a problem with the input we should make the error as soon as possible

popde/density_estimate.py
```diff
@@ -167,6 +167,67 @@ def evaluate_scipy(self, points):
...
+
+        # Create the contour plot
+        zz = z.reshape(xx.shape)
+        fig = utils_plot.simple2Dplot(xx, yy, zz, xlabel=dim1, ylabel=dim2, title='2D Contour Plot of KDE for {dim1} and {dim2} (Sliced along {slice_dims})')
+        if file_name is not None:
+            fig.savefig(file_name)
+        return fig
+
+
+
```
tdent, line 230: 2 blank lines .. 
```suggestion
```

popde/density_estimate.py
```diff
@@ -167,6 +167,67 @@ def evaluate_scipy(self, points):
         return density_values
 
 
+    def plot_2d_contour(self, dim1, dim2, slice_dims=None, slice_values=None, file_name=None, num_points=100, **kwargs):
```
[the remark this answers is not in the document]
JAMSADIQ: done

popde/density_estimate.py
```diff
@@ -167,6 +167,67 @@ def evaluate_scipy(self, points):
...
+            data = np.random.multivariate_normal(mean, covariance_matrix, num_points)
+            parameter = ['m1', 'm2', 'Mc']
+            kde = SimpleKernelDensityEstimation(sample,  dim_names=parameter)
+            # Plot a 2D contour with a slice along the 'z' dimensios
+            fig = kde.plot_2d_contour(parameter[0],parameter[1], 
+                    slice_dims=[parameter[2]], slice_values=[0], num_points=100)
+        """
+        if dim1 not in self.dim_names or dim2 not in self.dim_names:
+            raise ValueError("Invalid dimension names")
+        #get the samples along the dimensions
```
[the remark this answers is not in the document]

popde/density_estimate.py

tdent:
I think there is 1 suggestion still to be looked at (the plot title)

popde/density_estimate.py
```diff
@@ -167,6 +167,67 @@ def evaluate_scipy(self, points):
...
+            for slice_dim, slice_value in zip(slice_dims, slice_values):
+                slice_idx = self.dim_names.index(slice_dim)
+                positions = np.insert(positions, slice_idx, slice_value, axis=1)
+
+        # Evaluate the KDE at the grid points
+        z = self.evaluate(positions)
+
+        # Create the contour plot
+        zz = z.reshape(xx.shape)
+        fig = utils_plot.simple2Dplot(xx, yy, zz, xlabel=dim1, ylabel=dim2, title='2D Contour Plot of KDE for {dim1} and {dim2} (Sliced along {slice_dims})')
```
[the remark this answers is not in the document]
JAMSADIQ: done

tdent (approved):

PR #15, opened 2023-12-31
Input data processing / transformation

tdent:
Functions / methods to apply transformations to the input data, standardize the transformed data and rescale dimensions immediately before calculating the KDE.

### DIFF
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -1,61 +1,6 @@
 import numpy as np
 
 
-def apply_transf(column, fnc):
-    """
-    Apply a named transformation to a single array
-
-    'log' or 'ln' : take log of data
-    'stdize' : 
-    fnc convertible to float : multiply by constant
-    """
-    if fnc in ['log', 'ln']:
-        return np.log(column)
-    elif fnd == 'exp':
-        return np.exp(column)
-    elif fnc == 'stdize':
-        return column / np.std(column)
-    else:
-        try:  # Convert possible string to float
-            return column * float(fnc)
-        except:
-            raise ValueError(f'I got an unknown transformation function {fnc} !') 
-
-
-def transform_data(data, transf):
-    """
-    Cycle over data dimensions applying a named transformation to each 
-    """
-    assert len(data.shape) = 2  # only works for 'tabular' data
-    assert len(transf) = data.shape[1]
-    transf_data = np.zeros_like(data)
-    for (col, fnc) in zip(data.T, transf):  # iterate over columns
-        if fnc in ('None', 'none'):  # no-op
-             transf_data[:, dim] = col
-        else:
-             transf_data[:, dim] = apply_transf(col, fnc)
-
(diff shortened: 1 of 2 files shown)


popde/transform_utils.py
```diff
@@ -0,0 +1,59 @@
...
+    Apply a named transformation to a single array
+
+    'log' or 'ln' : take log of data
+    'stdize' :
+    fnc convertible to float : multiply by constant
+    """
+    column = np.asarray(column)
+    if fnc in ['log', 'ln']:
+        return np.log(column)
+    elif fnd == 'exp':
```
AnaLorenzoMedina, line 16: typo! fnd instead of fnc

popde/transform_utils.py
```diff
@@ -0,0 +1,59 @@
...
+def transform_data(data, transf):
+    """
+    Cycle over data dimensions applying a named transformation to each
+    """
+    assert len(data.shape) == 2  # only works for 'tabular' data
+    assert len(transf) == data.shape[1]
+    transf_data = np.zeros_like(data)
+    for (col, fnc) in zip(data.T, transf):  # iterate over columns
+        if fnc in ('None', 'none'):  # no-op
+             transf_data[:, dim] = col
```
AnaLorenzoMedina, line 36: 'dim' is not defined anywhere, maybe you wanted 'col' here?

popde/density_estimate.py
```diff
@@ -49,25 +57,65 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
         """
         if len(data.shape) != 2:
             raise ValueError("Data must have shape (n_samples, n_features).")
+        self.ndim = self.data.shape[1]
 
         self.data = np.asarray(data)
```
AnaLorenzoMedina, line 62: should this line go before line 58 (if len(data.shape) != 2:)? To transform data into an array before anything else, because lists for example don't have a list.shape method

popde/density_estimate.py
```diff
@@ -49,25 +57,65 @@ def __init__(self, data, backend='scipy', bandwidth=1., dim_names=None):
...
+        Transform, standardize and rescale input data into KDE-ready data
+        """
+        if self.transf is not None:
+            self.transf_data = transform_data(self.data, self.transf)
+        else:
+            self.transf_data = self.data
+
+        if self.stdize:
+            std_transf = ['stdize' for dim in self.ndim]
+            self.stds = numpy.std(self.transf_data, axis=0)  # record the stds
```
AnaLorenzoMedina, line 109: numpy is imported as np

popde/density_estimate.py
tdent: mmm, so we shouldn't be using a list as input in any case, if so there will be an error even earlier when doing data.shape, & I think it is 'correct' that we get an error.

AnaLorenzoMedina (approved):

PR #16, opened 2024-02-21
Awkde kdepy

JAMSADIQ:


JAMSADIQ:
Now code can be reviewd for script awkdepy.py

tdent:
So it looks like `awkdepy.py` is a module not a script .. that is what we want, but you should learn the difference. 

Concerning the organization of the code, I would suggest to call the new module `adaptive_kde.py` since that is what the objective is, the specific use of 'awkde' or 'kdepy' is not an essential aspect. 

Also, if we will now have an adaptive KDE module, can you move the 'old awkde' class  AdaptiveKDELeaveOneOutCrossValidation from density_estimate to adaptive_kde, as it belongs with the new adaptive code?

popde/awkdepy.py
```diff
@@ -0,0 +1,110 @@
+import numpy as np
+from density_estimate import SimpleKernelDensityEstimation
+from KDEpy.TreeKDE import TreeKDE
+
+class General_adaptiveKDE(SimpleKernelDensityEstimation):
```
tdent, line 5: Suggest just 'AdaptiveBwKDE' - I don't know if this is more or less general than other possibilities .. 
Also, is there a reason why this does not subclass VariableBwKDE?  As far as I can see the fit_KDEpy and evaluate_KDEpy methods there could just be used here once self.bandwith has been set

popde/awkdepy.py
```diff
@@ -0,0 +1,110 @@
...
+        -----------
+        kde_values : array-like
+            The KDE values at the data point positions.
+
+        Returns:
+        --------
+        inv_loc_bw : array-like
+            Inverse of the local bandwidth calculated using the provided function.
+        """
+        print("kdevals = ", kde_values)
```
tdent, line 60: Remove please

tdent:
Were you not able to get the code working by subclassing VariableBwKDE ? (That automatically already uses kdepy / TreeKDE so I think you would not have to set it here again.)

popde/awkdepy.py
JAMSADIQ: removed

popde/awkdepy.py
```diff
@@ -0,0 +1,110 @@
...
+    density_values = kde.evaluate(eval_pts)
+    density_values = density_values.reshape(XX.shape)
+    print(density_values)
+    """
+    def __init__(self, data, backend='KDEpy', bandwidth=1., dim_names=None, alpha=0.5,  input_transf=None, stdize=False, rescale=None):
+        # Additional parameter for the Wang & Wang formula
+        self.alpha = alpha
+        #self.bandwidth = bandwidth
+        #self.kde_data = data
+        super().__init__(data, input_transf, stdize, rescale,
```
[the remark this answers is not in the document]
JAMSADIQ: fixed this

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+    mean3, sigma3 = 0.0, 1.0
+    n_samples = 1000
+    rndgen = np.random.RandomState(seed=1)
+    sample1 = rndgen.normal(mean1, sigma1, size=n_samples)
+    sample2 = rndgen.normal(mean2, sigma2, size=n_samples)
+    sample3 = rndgen.normal(mean3, sigma3, size=n_samples)
+    sample = np.column_stack((sample1, sample2, sample3)) # shape is (n_points, n_features)
+    # Create and fit the adaptive KDE note backend must be set correctly
+    kde = AdaptiveBwKDE(sample, backend='awKDEpy', dim_names=['x', 'y', 'z'], alpha=0.5, input_transf=None)
+    print("kde=", kde)
```
tdent, line 23: again I don't think people will want to see the output of this print statement

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+    density_values = density_values.reshape(XX.shape)
+    print(density_values)
+    """
+    def __init__(self, data, backend='KDEpy', bandwidth=1., dim_names=None, alpha=0.5,  input_transf=None, stdize=False, rescale=None):
+        # Additional parameter for the Wang & Wang formula
+        self.alpha = alpha
+        super().__init__(data, input_transf, stdize, rescale,
+                         backend, bandwidth, dim_names)  
+
+    def _calc_local_bandwidth(self, kde_values):
```
tdent, line 44: you have called this calc local bandwidth, but it doesn't calculate local bw, it calculates the inverse bw factor which we apply to the original global bw. 
Change the name please

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+
+        Parameters:
+        -----------
+        kde_values : array-like
+            The KDE values at the data point positions.
+
+        Returns:
+        --------
+        inv_loc_bw : array-like
+            Inverse of the local bandwidth calculated using the provided function.
```
tdent, line 56: no, inv_loc_bw is not the inverse of the local bw, it is the inverse factor we apply to the global bw.  Also 'calculated using the provided function' does not give the reader any information.  If you want to explain _which_ function you could give a reference to an equation in the Wang paper

tdent:
Have you tested whether this class actually gives a reasonable result in making an adaptive KDE?

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+        else:
+            per_point_bandwidths = np.ones_like(local_bandwidths) / self.bandwidth
+
+        return per_point_bandwidths
+
+    def fit_awKDEpy(self):
+        """
+        Fit the adaptive KDE
+        """
+        pilot_kde = TreeKDE(bw=self.bandwidth).fit(self.kde_data)
```
tdent, line 93: do you still need to fit a pilot KDE here? Doesn't `VariableBwKDEPy.__init__` already use the fit() method, and so the output of this is already present and could be used as the pilot KDE?

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+        inv_loc_bw : array-like
+            Inverse of the local bandwidth calculated using the provided function.
+        """
+        g = np.exp(np.sum(np.log(kde_values)) / len(kde_values))
+        inv_loc_bw = (kde_values / g) ** self.alpha
+        return inv_loc_bw
+
+    def calculate_per_point_bandwidths(self, pilot_values, adaptive=True):
+        """
+        Calculate per-point bandwidths using the Wang & Wang formula.
```
tdent, line 64: rather than 'the Wang & Wang formula' give an equation reference and the paper reference. Someone else reading the code won't know what 'the Wang & Wang formula' means

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+        Returns:
+        --------
+        inv_loc_bw : array-like
+            Inverse of the local bandwidth calculated using the provided function.
+        """
+        g = np.exp(np.sum(np.log(kde_values)) / len(kde_values))
+        inv_loc_bw = (kde_values / g) ** self.alpha
+        return inv_loc_bw
+
+    def calculate_per_point_bandwidths(self, pilot_values, adaptive=True):
```
tdent, line 62: why do we need an 'adaptive=True' kwarg here?  If someone wants a non-adaptive KDE they can use another class, or they could set alpha=0. I don't see the need for the extra complexity in this method

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+        adaptive : bool, optional (default=True)
+            If True, use adaptive bandwidth; if False, use fixed bandwidth.
+
+        Returns:
+        --------
+        per_point_bandwidths : array-like
+            Per-point bandwidths calculated using the Wang & Wang formula.
+        """
+        # Calculate the local bandwidth using the provided function
+        local_bandwidths = self._calc_local_bandwidth(pilot_values)
```
tdent, line 79: the variable should be called something like 'inverse bw factors', as it does _not_ return local bws

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # Use the local bandwidths to calculate per-point bandwidths
+        if adaptive:
+            per_point_bandwidths = self.bandwidth * local_bandwidths
```
tdent, line 83: as far as I understand the formula, the pointwise bws should be the original global bw _divided by_ the 'local_bandwidths' factors.  Please check this with the paper

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+        else:
+            per_point_bandwidths = np.ones_like(local_bandwidths) / self.bandwidth
```
tdent, line 85: here, you are _dividing_ by self.bandwidth.  This doesn't seem consistent.  If you want `per_point_bandwidths` to behave like the global bw variable, it should be * not /

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+        return per_point_bandwidths
+
+    def fit_awKDEpy(self):
+        """
+        Fit the adaptive KDE
+        """
+        pilot_kde = TreeKDE(bw=self.bandwidth).fit(self.kde_data)
+        pilot_values = pilot_kde.evaluate(self.kde_data)
+        # Calculate per-point bandwidths
+        per_point_bandwidths = self.calculate_per_point_bandwidths(pilot_values)
```
tdent, line 96: can't you reassign self.bandwidth to be equal to per_point_bandwidths to convert from a global bw to the local one?

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # Update the KDE with per-point bandwidths
+        self.kernel_estimate = TreeKDE(bw=self.bandwidth).fit(self.kde_data)
```
tdent, line 99: so if you reassign self.bandwidth to be the array of per-point bandwidths, you can use self.fit() to make the adaptive KDE

popde/adaptive_kde.py
JAMSADIQ: That is not clear to me how pilot kde can be obtained from superclass here

tdent: well, look at the VariableBwKDEPy.__init__ method.  It does `self.fit()`.  That calls `fit_method` which here is `fit_KDEpy`, which sets `self.kernel_estimate` to be a TreeKDE object (line 193 of density_estimate.py).  So then you can use `self.kernel_estimate`, no?

In general, you should try to use methods or attributes that already exist to avoid duplication.

tdent: I am assuming you read the code for the superclass, if you see a specific issue there can you mention it, otherwise I can't tell what may not be clear.

popde/adaptive_kde.py

popde/adaptive_kde.py
```diff
@@ -0,0 +1,191 @@
...
+        -----------
+        kde_values : array-like
+            The KDE values at the data point positions.
+
+        Returns:
+        --------
+        inv_loc_bw : array-like
+            Inverse of the local bandwidth calculated using the provided function.
+        """
+        g = np.exp(np.sum(np.log(kde_values)) / len(kde_values))
```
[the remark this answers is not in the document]
JAMSADIQ: fixed

popde/adaptive_kde.py
JAMSADIQ: changes done

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: fixed

popde/adaptive_kde.py
JAMSADIQ: changed,

popde/adaptive_kde.py
JAMSADIQ: removed else

popde/adaptive_kde.py
JAMSADIQ: fixed

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: corrected

popde/adaptive_kde.py
JAMSADIQ: Using 
 `pilot_kde = self.fit()`
I got an error:

> density_estimate.py", line 126, in fit
>     fit_method()
> RecursionError: maximum recursion depth exceeded

JAMSADIQ: work in progress

popde/adaptive_kde.py
JAMSADIQ: fixed

popde/adaptive_kde.py
tdent: ah, so what problem might this indicate?

tdent: so I did not say you had to call self.fit() again, because that is already called in the __ini__ method of the parent class.  I said you could use self.kernel_estimate ..

tdent: in general, you can put in print statements to see what is going wrong, eg you could put in `print(fit_method)` or `print(self.backend)`

tdent:
@JAMSADIQ I was thinking that the `set_bandwidth` method of the KDEpy class could be used for the adaptive fitting.

popde/adaptive_kde.py
JAMSADIQ: no, I did not push the failed script on PR. Also self.estimate give error,  
work in progress

JAMSADIQ: issue maybe in setting  backened option.

JAMSADIQ: this commit did not work but commit the change.
fix required

popde/density_estimate.py
```diff
@@ -1,5 +1,5 @@
 import numpy as np
-import transform_utils as transf
+#import transform_utils as transf
```
tdent, line 2: this should not be commented out .. ?

tdent:
Should I look at this?  If there is a problem, can you describe what it is (ie in what case do you get an error and what is the error?)

popde/adaptive_kde.py
```diff
@@ -0,0 +1,206 @@
...
+                g is local factor given in local_bandwidth_factor
+        """
+        # Calculate the inverse of local bandwidth using the provided function
+        local_bandwidths = self._local_bandwidth_factor(pilot_values)
+
+        # Use the local bandwidths to calculate per-point bandwidths
+        self.bandwidth = self.bandwidth / local_bandwidths
+
+
+    def fit_awKDEpy(self):
```
tdent, line 91: yes, you do not need this method, the adaptive kde calculation of bandwidth can be done within the __ini__ method of this class and then you can use the reassign_bandwidth method of the parent class to set the loca bandwidth and redo the fit.

popde/adaptive_kde.py
```diff
@@ -0,0 +1,206 @@
...
+        kde_values : array-like
+            The KDE values at the data point positions.
+
+        Returns:
+        --------
+        loc_bw_factor : array-like
+           local bandwidth factor for adaptive 
+           bandwidth.
+        """
+        from scipy.stats import gmean
```
tdent, line 59: this should be imported at the start of the module to avoid that we do the import repeatedly (eg if we were optimizing the adaptive parameters)

popde/adaptive_kde.py
```diff
@@ -0,0 +1,206 @@
...
+        Parameters:
+        -----------
+        pilot_values : array-like
+            The pilot KDE values at the data point positions.
+
+        Returns:
+        --------
+        per_point_bandwidths : array-like
+            Per-point bandwidths calculated using the Wang & Wang formula.
+            bw_i =  global_bw * local_bw_factor
```
tdent, line 78: you do not need to have all this extra explanation here, just give the number of the equation in the paper in the reference at the beginning

popde/adaptive_kde.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        from scipy.stats import gmean
+        #geometric mean of kde values
+        g = gmean(kde_values)
+        loc_bw_factor = (kde_values / g) ** self.alpha
+        return loc_bw_factor
+
+    def calculate_per_point_bandwidths(self, pilot_values):
+        """
+        Calculate per-point bandwidths using the Wang & Wang formula in
```
tdent, line 67: if you cite the paper it is unnecessary to also use the slang 'the Wang & Wang formula' ...
```suggestion
        Calculate per-point bandwidths according to
```

JAMSADIQ: fixed

popde/adaptive_kde.py

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: removed this and added adaptive bandwidth in init.

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+    print(density_values)
+    """
+    def __init__(self, data, backend='KDEpy', bandwidth=1., dim_names=None, alpha=0.5,  input_transf=None, stdize=False, rescale=None):
+        # Additional parameter for the Wang & Wang formula
+        self.alpha = alpha
+        super().__init__(data, input_transf, stdize, rescale,
+                         backend, bandwidth, dim_names) 
+
+        #Adaptive bandwidth: compute pilot kdevals
+        pilot_values = self.evaluate(self.kde_data)
```
tdent, line 43: ok, since it may be useful to have these available can you assign this to self.pilot_values instead

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        #geometric mean of kde values using scipy
+        g = gmean(kde_values)
+        loc_bw_factor = (kde_values / g) ** self.alpha
+        return loc_bw_factor
+
+    def calculate_per_point_bandwidths(self, pilot_values):
+        """
+        Calculate per-point bandwidths according to
+        B. Wang and X. Wang, 2007, DOI: 10.1214/154957804100000000.
```
tdent, line 71: Since the actual formula is only used in the previous method, and this function is just dividing the existing bw by the factor, remove the citation. 
Also, the method doesn't just _calculate_ the local bws, it also sets self.bandwidth to a new value.
So I would rather name the method eg 'set_per_point_bandwidths'

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+    XX, YY, ZZ = np.meshgrid(xgrid, ygrid, zgrid)
+    eval_pts = np.column_stack((XX.flatten(), YY.flatten(), ZZ.flatten()))
+
+    # Evaluate the KDE at the grid points
+    density_values = kde.evaluate(eval_pts)
+    density_values = density_values.reshape(XX.shape)
+    print(density_values)
+    """
+    def __init__(self, data, backend='KDEpy', bandwidth=1., dim_names=None, alpha=0.5,  input_transf=None, stdize=False, rescale=None):
+        # Additional parameter for the Wang & Wang formula
```
tdent, line 37: explanation of 'alpha' should be in the docstring before the example, not here.
```suggestion
```

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+        self.alpha = alpha
```
tdent, line 38: we may want to keep track of the global bandwidth, so add a line `self.global_bandwidth = bandwidth`

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+    def calculate_per_point_bandwidths(self, pilot_values):
+        """
+        Calculate per-point bandwidths according to
+        B. Wang and X. Wang, 2007, DOI: 10.1214/154957804100000000.
+        Parameters:
+        -----------
+        pilot_values : array-like
+            The pilot KDE values at the data point positions.
+
+        Returns:
```
tdent, line 77: This method doesn't return anything!

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+        --------
+        per_point_bandwidths : array-like
+        """
+        # Calculate the inverse of local bandwidth using the provided function
```
tdent, line 81: Comment is too long and inaccurate anyway (we're calculating the inverse bandwidth _factor_, not the invers of the bandwidths)
```suggestion
        # Calculate adaptive inverse bandwidth factors
```

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+        local_bandwidths = self._local_bandwidth_factor(pilot_values)
```
tdent, line 82: this variable name is misleading.  But I don't see why you need to define a variable anyway as it's only used once

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Use the local bandwidths to calculate per-point bandwidths
+        self.bandwidth = self.bandwidth / local_bandwidths
```
tdent, line 85: Can you please look at https://github.com/tdent/pop-de/blob/main/popde/density_estimate.py#L134 
This is a method that sets self.bandwidth _and_ re-does the fit with the new bandwidth, which your code doesn't seem to do.

tdent: So can you use the set_bandwidth method here?

popde/adaptive_kde.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+    def evaluate(self, points):
```
tdent, line 88: I don't see why you need a new evaluate method here. The class already inherits the evaluate method from VariableBwKDEPy and that will be evaluate_KDEpy as here https://github.com/tdent/pop-de/blob/main/popde/density_estimate.py#L260 
So this method seems exactly the same as what was already coded, I don't think you need it

JAMSADIQ: removed this method

popde/adaptive_kde.py
JAMSADIQ: fixed

popde/adaptive_kde.py
JAMSADIQ: removed this line

popde/adaptive_kde.py
JAMSADIQ: change this

popde/adaptive_kde.py
JAMSADIQ: fixed this method

popde/adaptive_kde.py
JAMSADIQ: fixed

popde/adaptive_kde.py
JAMSADIQ: done both

popde/adaptive_kde.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _local_bandwidth_factor(self, kde_values):
+        """
+        Calculate local bandwidth factor using expression in
+        B. Wang and X. Wang, 2007, DOI: 10.1214/154957804100000000.
+
+        Parameters:
+        -----------
+        kde_values : array-like
+            The KDE values at the data point positions.
```
tdent, line 58: better to say 'pilot KDE values' as that is how we use it

popde/adaptive_kde.py
```diff
@@ -0,0 +1,172 @@
...
+
+        Returns:
+        --------
+        loc_bw_factor : array-like
+           local bandwidth factor for adaptive 
+           bandwidth.
+        """
+        #geometric mean of kde values using scipy
```
tdent, line 66:
```suggestion
        # Geometric mean of pilot kde values
```
Since gmean is clearly imported from scipy, no need to mention it here

popde/adaptive_kde.py
```diff
@@ -0,0 +1,172 @@
...
+        Parameters:
+        -----------
+        pilot_values : array-like
+            The pilot KDE values at the data point positions.
+        """
+        # Calculate adaptive inverse bandwidth factors
+        # and calculate per-point bandwidths
+        self.bandwidth /= self._local_bandwidth_factor(pilot_values)
+        #change bandwidth and re-initialize KDE 
+        self.set_bandwidth(self.bandwidth)
```
tdent, line 84: I think these two lines can be replaced by one; also better to refer to self.global_bandwidth if we want specifically to use the global bw in this formula, to remove the risk of applying the adaptive factor multiple times to self.bandwidth
```
self.set_bandwidth(
    self.global_bandwidth / self._local_bandwidth_factor(pilot_values)
)
```

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: fixed

popde/adaptive_kde.py
JAMSADIQ: done

JAMSADIQ:
initiate

tdent (approved):

PR #19, opened 2024-03-25
dynamicadaptiveparameters

JAMSADIQ:
for issue 17:  add methods which allow the values to be changed and also redo the necessary adaptive fitting steps

### DIFF
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -91,11 +91,9 @@ class AdaptiveBwKDE(VariableBwKDEPy):
             new_alpha (float): The new value for the adaptive parameter alpha.
         """
         self.alpha = new_alpha
-        # compute pilot_values
-        self.pilot_values = self.evaluate(self.kde_data)
-
-        #update bandwidth with new alpha
+        #update bandwidth with new_alpha
         self.set_per_point_bandwidth(self.pilot_values)
+       
 
 
     def set_adaptive_parameter(self, new_alpha, new_global_bw):
@@ -108,15 +106,9 @@ class AdaptiveBwKDE(VariableBwKDEPy):
             new_alpha (float): The new value for the adaptive parameter alpha.
             new_global_bw (float): The new value for the global bandwidth.
         """
-        self.set_alpha(new_alpha)
-        # use new global bandwith and fit it
         self.set_bandwidth(new_global_bw)
-
-        # Adaptive bandwidth: compute pilot kdevals
-        self.pilot_values = self.evaluate(self.kde_data)
-
-        # Calculate per-point bandwidths to reassign self.bandwidth and re-initialize KDE
-        self.set_per_point_bandwidth(self.pilot_values)
+        #calculate per-point bandwidths using 
+        self.set_alpha(new_alpha)
 
 
 class AdaptiveKDELeaveOneOutCrossValidation():
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -82,6 +82,33 @@ class AdaptiveBwKDE(VariableBwKDEPy):
         self.set_bandwidth(
             self.global_bandwidth / self._local_bandwidth_factor(pilot_values)
         )
+    
+     def set_alpha(self, new_alpha):
+        """
+        Set the adaptive parameter alpha to a new value.
+
+        Parameter:
+            new_alpha (float): The new value for the adaptive parameter alpha.
+        """
+        self.alpha = new_alpha
+
+    def set_adaptive_parameter(self, new_alpha, new_global_bw):
+        """
+        This method updates the adaptive parameter alpha and the global bandwidth, then performs adaptive calculations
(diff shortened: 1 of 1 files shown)


popde/adaptive_kde.py
```diff
@@ -82,6 +82,33 @@ def set_per_point_bandwidth(self, pilot_values):
...
         )
+    
+     def set_alpha(self, new_alpha):
+        """
+        Set the adaptive parameter alpha to a new value.
+
+        Parameter:
+            new_alpha (float): The new value for the adaptive parameter alpha.
+        """
+        self.alpha = new_alpha
```
tdent, line 93: here, don't we also want to redo the calculation of the bandwidths to account for the change in alpha?  Otherwise we would have an object where the bandwidths are still the old ones

popde/adaptive_kde.py
```diff
@@ -82,6 +82,33 @@ def set_per_point_bandwidth(self, pilot_values):
...
+
+    def set_adaptive_parameter(self, new_alpha, new_global_bw):
+        """
+        This method updates the adaptive parameter alpha and the global bandwidth, then performs adaptive calculations
```
tdent, line 97:
```suggestion
        Update the adaptive parameter alpha and the global bandwidth, then perform adaptive calculations
```

popde/adaptive_kde.py
tdent: (Specifically, this method would be used if only alpha is changed but not the global bw, so the pilot values would remain the same.)

popde/adaptive_kde.py
```diff
@@ -82,6 +82,33 @@ def set_per_point_bandwidth(self, pilot_values):
...
+
+        Parameters:
+            new_alpha (float): The new value for the adaptive parameter alpha.
+            new_global_bw (float): The new value for the global bandwidth.
+        """
+        self.set_alpha(new_alpha)
+        self.global_bandwidth = new_global_bw
+
+        # Adaptive bandwidth: compute pilot kdevals
+        self.pilot_values = self.evaluate(self.kde_data)
```
tdent, line 108: here, what bandwidth will the class use to do self.evaluate?  Since you have only changed the self.global_bandwidth attribute but no new fitting has been done, I think self.evaluate will still use the old bandwidths. You will need to e.g. use set_bandwidth with the new global bw to force a new fit.

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: update this, work in progress

popde/adaptive_kde.py
JAMSADIQ: fixed,

popde/adaptive_kde.py
JAMSADIQ: done

JAMSADIQ:
ok to review it.

popde/adaptive_kde.py
```diff
@@ -82,6 +82,41 @@ def set_per_point_bandwidth(self, pilot_values):
...
+    def set_alpha(self, new_alpha):
+        """
+        Set the adaptive parameter alpha to a new value.
+
+        Parameter:
+            new_alpha (float): The new value for the adaptive parameter alpha.
+        """
+        self.alpha = new_alpha
+        # compute pilot_values
+        self.pilot_values = self.evaluate(self.kde_data)
```
tdent, line 95: do the pilot values actually need to be recalculated here?  I think they do not depend on alpha (only on global bw)

JAMSADIQ: They are needed here to update bandwidth

tdent: yes, I know you need to _use_ pilot values, but I am saying that `self.pilot_values` already exists and you do not have to recalculate them

tdent:
Hopefully these methods still work OK and give sensible results?  (eg you could check by putting in print statements to check the values of self.bandwidth)

popde/adaptive_kde.py
```diff
@@ -74,14 +74,43 @@ def set_per_point_bandwidth(self, pilot_values):
...
+        then reassign local bandwidths and re-initialize the KDE.
+
+        Parameters
+        ----------
+        new_alpha : float
+            The new value for the adaptive parameter alpha.
+        new_global_bw : float
+            The new value for the global bandwidth.
+        """
+        self.set_bandwidth(new_global_bw)
```
tdent, line 111: Does this recalculate self.pilot_values ? As bw changes, the pilot values _do_ need to be recalculated here

JAMSADIQ:
alpha, bandwidth do get updated when this function "kde.set_adaptive_parameter(newalp, newbw)" is called but printing kde.bandwidth returns just updated global bandwidth but not an array.

tdent (approved):

tdent:
I fixed the docstring and comment formats but it would be good if you remember to use numpy style comments / docstrings.

Issue #20, opened 2024-04-08
Transform KDE in kde_data variables back to original variables


tdent:
At present the KDE values are in variables (coordinates) determined by `self.kde_data` : i.e. the KDE is a density over a space defined in specific coordinates.  In general these will not be physically useful or meaningful units. 

So, make the required inverse transformations in the density to convert it to the units of `self.transf_data` ('transformed data units') and then also to `self.data` ('original data units'). 

This is probably best done by adding new methods to the basic density estimate class ('SimpleKernelDensityEstimation') which take an array of KDE values and transform it from `kde_data` units to the desired ones.

tdent:
Closed by #21

PR #21, opened 2024-04-19
Transform data

JAMSADIQ:
initiated a method to work on issue #20

### DIFF
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -112,13 +112,12 @@ class SimpleKernelDensityEstimation:
             self.stds = np.std(self.transf_data, axis=0)  # record the stds
             self.std_data = transf.transform_data(self.transf_data, std_transf)
         else:
-            self.stds = None
             self.std_data = self.transf_data
 
         if self.rescale is not None:
-            print("resclae", self.rescale)
-            quit()
+            print("self tests tran, rescale", self.input_transf, self.rescale)
             self.kde_data = transf.transform_data(self.std_data, self.rescale)
+            print("rescaling done")
         else:
             self.kde_data = self.std_data
 
@@ -148,12 +147,12 @@ class SimpleKernelDensityEstimation:
         if self.stdize:
             std_points = transf.transform_data(transf_points, 1.0/self.stds)
         else:
-            std_values_for_points = None
             std_points = transf_points
 
         # Rescale each parameter
         if self.rescale is not None:
-            transf_data = transf.transform_data(std_points, 1.0/self.rescale)
+            transf_data = transf.transform_data(std_points, self.rescale)
+            
         else:
             transf_data = std_points
 
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -108,7 +108,7 @@ class SimpleKernelDensityEstimation:
             self.transf_data = self.data
 
         if self.stdize:
-            std_transf = ['stdize'] *self.ndim
+            std_transf = ['stdize'] * self.ndim
             self.stds = np.std(self.transf_data, axis=0)  # record the stds
             self.std_data = transf.transform_data(self.transf_data, std_transf)
         else:
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -126,12 +126,13 @@ class SimpleKernelDensityEstimation:
         Apply specified transformations to KDE values for any variable
 
         Parameters:
-        eval_values: np.ndarray, the evaluation points (N, D) where N is number of points and D is dimensions
+        eval_values: np.ndarray, the evaluation points 
+        with shape [N_samples, N_features] 
         kde_values : numpy.ndarray
             The KDE values to be transformed.
-            It should be an array with dimensions (N_samples, N_dimensions),
+            It should be an array with dimensions (N_samples, N_features),
             where N_samples is the number of samples 
-            and N_dimensions is the number of dimensions of the KDE.
+            and N_features is the number of dimensions of the KDE.
 
         transformation_options : list of str
             A list containing transformation options for 
@@ -142,37 +143,29 @@ class SimpleKernelDensityEstimation:
 
         Returns:
         numpy.ndarray
-            The transformed KDE values.
+            The transformed KDE values. (Jacobian of transformation)
 
         Raises:
         ValueError: If an invalid transformation option is provided.
         apply tranformation on kde values
         """
-        # Validate the dimensions assuming eval_values have n_samples, n_feature
+        # Validate the dimensions assuming eval_values have n_samples, n_features
         num_points, num_dims = eval_values.shape
         if len(transformation_options) != num_dims:
             raise ValueError("The transformation list length must match the number of dimensions in eval_vals")
        
-        #get variables grid values separately for Jacobian: This can be problematic
+        #get variables grid values separately for Jacobian: can be problematic
         eval_pts = []
         for dim in range(num_dims):
-            print(dim)
             variable_values = eval_values[:, dim]
             # Determine the unique values and their order
             unique_values = np.unique(variable_values)
-            print(unique_values) 
 
             # Determine the grid shape for the current dimension
-            print(len(np.unique(eval_values[:, 0])))
-            print(len(np.unique(eval_values[:, 1])))
-            print(len(np.unique(eval_values[:, 2])))
-            aa = [len(np.unique(eval_values[:, i])) for i in range(num_dims)]
-            
-            shape = tuple(aa)#(len(np.unique(eval_values[:, i])) for i in range(num_dims))
+            shape = tuple([len(np.unique(eval_values[:, i])) for i in range(num_dims)])
 
             # Reshape the dimension values to the determined grid shape
             grid = variable_values.reshape(shape)
-            print(grid.shape)
             eval_pts.append(grid)
 
         new_kde_values = np.copy(kde_values)#avoid modifying the original array
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -136,6 +136,7 @@ class SimpleKernelDensityEstimation:
         kde_vals : array-like, shape (n_samples,)
         The KDE values adjusted by the Jacobian of the transformations.
         """
+        # Initial transformation
         if self.input_transf is not None:
             transf_points = transf.transform_data(points, self.input_transf)
         else:
@@ -143,11 +144,11 @@ class SimpleKernelDensityEstimation:
 
         # Divide each parameter by the std of the training data
         if self.stdize:
-            std_points = transf.transform_data(transf_points, 1.0/self.stds)
+            std_points = transf.transform_data(transf_points, 1. / self.stds)
         else:
             std_points = transf_points
 
-        # Rescale each parameter
+        # Rescaling
         if self.rescale is not None:
             transf_data = transf.transform_data(std_points, self.rescale)       
         else:
@@ -156,28 +157,22 @@ class SimpleKernelDensityEstimation:
         # Evaluate kde on transform points
         kde_vals = self.evaluate(transf_data)
 
-        # Jacobian of transform
+        # Jacobian of transforms for each dimension
         for i, option in enumerate(self.input_transf):
             if option in['log', 'ln']:
-                input_Jacobian = 1.0 / points[:, i]
+                input_Jacobian = 1. / points[:, i]
             elif option == 'exp':
                 input_Jacobian = np.exp(points[:, i])
             elif option in ('none', 'None'):
-                input_Jacobian = 1.0 
+                input_Jacobian = 1. 
             else:
                 raise ValueError(f"Invalid transformation option at index {i}: {option}")
 
             # Jacobian for training data standardization
-            if self.stdize:
-                std_Jacobian = self.stds[i]
-            else:
-                std_Jacobian = 1.0
+            std_Jacobian = self.stds[i] if self.stdize else 1.
 
             # Jacobian for rescaling factor
-            if self.rescale is not None:
-                rescale_Jacobian = self.rescale[i]
-            else:
-                rescale_Jacobian = 1.0
+            rescale_Jacobian = self.rescale[i] if self.rescale is not None else 1.
 
             kde_vals *= input_Jacobian * std_Jacobian * rescale_Jacobian 
 
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -123,73 +123,58 @@ class SimpleKernelDensityEstimation:
 
     def evaluate_with_transf(self, points):
         """
-        Apply specified transformations to KDE values for any variable
+        Transforms the evaluate points based on specified transformations
+        , evaluates the kernel density estimate (KDE) on the transformed points
+         and adjusts the KDE values using the Jacobian of the transformations.
 
         Parameters:
-        eval_values: np.ndarray, the evaluation points 
-        with shape [N_samples, N_features] 
-        kde_values : numpy.ndarray
-            The KDE values to be transformed.
-            It should be an array with dimensions (N_samples, N_features),
-            where N_samples is the number of samples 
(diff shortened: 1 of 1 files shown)


tdent (changes requested):
We discussed that the code is not really ready for review. Please request again when it is ...

JAMSADIQ:
added some changes that seems to give reasonable output but maybe some ideas are needed to make it more efficient.

JAMSADIQ:
Can you review the current updated code

tdent (dismissed):

tdent:
The previous requested change from April was just to set up a version that can be reviewed, so I will start reviewing the new version now ..

popde/density_estimate.py
```diff
@@ -120,6 +120,70 @@ def prepare_data(self):
         else:
             self.kde_data = self.std_data
 
+
+    def evaluate_with_transf(self, points):
+        """
+        Transforms the evaluate points based on specified transformations
+        , evaluates the kernel density estimate (KDE) on the transformed points
+         and adjusts the KDE values using the Jacobian of the transformations.
```
tdent, line 128:
```suggestion
        Transforms the input points in the same way as the KDE training data,
        evaluates the KDE on the transformed points and adjusts the KDE
        values by the Jacobian of the transformations.
```

popde/density_estimate.py
```diff
@@ -120,6 +120,70 @@ def prepare_data(self):
...
+            std_points = transf.transform_data(transf_points, std_transf)
+        else:
+            std_values_for_points = None
+            std_points = transf_points
+
+        #rescale
+        if self.rescale is not None:
+            transf_data = transf.transform_data(std_points, self.rescale)
+        else:
+            transf_data = std_points
```
tdent, line 158: This looks OK, but there should be some way of combining this set of transformations on the evaluate points with the prepare_data() method .. eg one could change prepare_data() into a method that takes some set of data points and outputs the transformed points and the stds (if standardization is done). 

Then this basic 'prepare some data' method could be called by other methods that either prepare the input training points or the evaluate points.

popde/density_estimate.py
```diff
@@ -245,6 +309,7 @@ def plot_2d_contour(self, dim1, dim2, slice_dims=None, slice_values=None, num_po
         return fig
 
 
+
```
tdent, line 312:
```suggestion
```

popde/density_estimate.py
```diff
@@ -120,6 +120,70 @@ def prepare_data(self):
...
+            transf_data = transf.transform_data(std_points, self.rescale)
+        else:
+            transf_data = std_points
+
+        #Evaluate kde on transform points
+        kde_vals = self.evaluate(transf_data)
+
+        num_points, num_dims = transf_data.shape
+        if len(self.input_transf) != num_dims:
+            raise ValueError("The transformation list length must match the number of dimensions in eval_vals")
```
tdent, line 165: I am not seeing why we need this, or why it is correct .. self.input_transf can be None, and if it is not None we have the check_dimensionality method to determine if the length is the right one. In any case the evaluate method should fail if transf_data has the wrong number of dimensions.

popde/density_estimate.py
```diff
@@ -120,6 +120,70 @@ def prepare_data(self):
...
+      
+        #Transform transf_data back to original
+        back_original_points = transf.reverse_transform(transf_data, self.input_transf, std_values_for_points, self.rescale)
```
tdent, line 168: is there a reason why you can't use the existing `points` ?

popde/density_estimate.py
```diff
@@ -120,6 +120,68 @@ def prepare_data(self):
...
+        #Evaluate kde on transform points
+        kde_vals = self.evaluate(transf_data)
+
+        num_points, num_dims = transf_data.shape
+        if len(self.input_transf) != num_dims:
+            raise ValueError("The transformation list length must match the number of dimensions in eval_vals")
+      
+        #get Jacobians using points in original units
+        for i, option in enumerate(self.input_transf):
+            print(i, type(option))
```
tdent, line 169: you can put print statements in for testing, but please remove from the code PR

popde/density_estimate.py
```diff
@@ -120,6 +120,68 @@ def prepare_data(self):
...
+            if option in['log', 'ln']:
+                print("option is ", option)
+                # Apply log transformation for that variable
+                kde_vals *= 1.0 / points[:, i]
+            elif option == 'exp':
+                # Apply exponential transformation
+                kde_vals *= np.exp(points[:, i])
+            elif option =='none':
```
tdent, line 177: you should allow for both 'none' and 'None' - see https://github.com/tdent/pop-de/blob/main/popde/transform_utils.py#L35

popde/density_estimate.py
```diff
@@ -120,6 +120,68 @@ def prepare_data(self):
...
+                print("no need for Jacobian")
+            else:
+                raise ValueError(f"Invalid transformation option at index {i}: {option}")
+
```
tdent, line 181: so are you going to also include the standardization and rescaling Jacobians ?

JAMSADIQ: This I am not understanding.

tdent: in the `prepare_data` method there are 3 possible operations on the input parameter values, a transformation function, a standardization (ie division by the std) and a rescaling. So far this PR only accounts for the transformation function and not the other two operations

tdent:
The input transformations can be written 
```
x -> y -> z -> w
```
where x are original input points, y are transformed (log/exp) data, z are standardized (std = 1) and w are rescaled points. We have
```
w = c_1 c_2 [ln, exp] x
```
where c_1 is the standardizing factor [which we set to 1/std(y)] and c_2 is an optional rescaling factor. So for the differentials 
```
dw = c_1 c_2 [1/x, e^x] dx
```
for the log / exp case resp. 

The KDE over w is a probability density p(w), we write 
```
dp = p(w) dw = p(x) dx
p(x) = (dw / dx) p(w)
```
where (dw / dx) is the Jacobian.  This is equal to `c_1 c_2 [1/x, e^x]` for [log, exp].

popde/density_estimate.py
```diff
@@ -120,6 +120,76 @@ def prepare_data(self):
...
+        kde_vals = self.evaluate(transf_data)
+
+        # Jacobian of transform
+        for i, option in enumerate(self.input_transf):
+            if option in['log', 'ln']:
+                input_Jacobian = 1.0 / points[:, i]
+            elif option == 'exp':
+                input_Jacobian = np.exp(points[:, i])
+            elif option in ('none', 'None'):
+                #print("no need for Jacobian")
```
tdent, line 168: remove print line .. 
```suggestion
```

popde/density_estimate.py
```diff
@@ -120,6 +120,76 @@ def prepare_data(self):
...
+                Jacobian_std_factor = 1.0
+
+            #rescale
+            if self.rescale is not None:
+                Jacobian_rescale_factor = self.rescale[i]
+            else:
+                Jacobian_rescale_factor = 1.0
+
+            Total_Jacobian =  input_Jacobian * Jacobian_std_factor * Jacobian_rescale_factor
+            #multiply KDE with Jacobian factor
```
tdent, line 186: comment not needed, obvious from reading the code
```suggestion
```

popde/density_estimate.py
```diff
@@ -120,6 +120,76 @@ def prepare_data(self):
...
+
+        # Jacobian of transform
+        for i, option in enumerate(self.input_transf):
+            if option in['log', 'ln']:
+                input_Jacobian = 1.0 / points[:, i]
+            elif option == 'exp':
+                input_Jacobian = np.exp(points[:, i])
+            elif option in ('none', 'None'):
+                #print("no need for Jacobian")
+                input_Jacobian  = points[:, i]
```
tdent, line 169: what is this doing ??

popde/density_estimate.py
```diff
@@ -120,6 +120,76 @@ def prepare_data(self):
...
+            else:
+                raise ValueError(f"Invalid transformation option at index {i}: {option}")
+
+            #stdardize 
+            if self.stdize is not None:
+                Jacobian_std_factor =  1.0/np.std(input_Jacobian)
```
tdent, line 175: what is this supposed to be?  The class instance already has an attribute recording the stds as you see from the prepare_data method

popde/density_estimate.py
```diff
@@ -120,6 +120,76 @@ def prepare_data(self):
...
+            #rescale
+            if self.rescale is not None:
+                Jacobian_rescale_factor = self.rescale[i]
+            else:
+                Jacobian_rescale_factor = 1.0
+
+            Total_Jacobian =  input_Jacobian * Jacobian_std_factor * Jacobian_rescale_factor
+            #multiply KDE with Jacobian factor
+            kde_vals   *= Total_Jacobian
+
```
tdent, line 188:
```suggestion
```

tdent:
@JAMSADIQ do you agree with the formulae I wrote?  I am still not seeing what you are implementing in the code

popde/density_estimate.py
JAMSADIQ: replace as this was a mistake

popde/density_estimate.py
```diff
@@ -120,6 +120,73 @@ def prepare_data(self):
...
+        Returns:
+        --------
+        kde_vals : array-like, shape (n_samples,)
+        The KDE values adjusted by the Jacobian of the transformations.
+        """
+        if self.input_transf is not None:
+            transf_points = transf.transform_data(points, self.input_transf)
+        else:
+            transf_points = points
+        #normalize
```
tdent, line 144:
```suggestion
        # Divide each parameter by the std of the training data
```

tdent: Note that we have to make _the same_ transformation on the evaluation points as the training points.  So, the standardization step divides each parameter by the std of the training data.  This is just a constant factor.  But then we have to divide the evaluation points by _the same factor_.  Not by the std of the evaluation points (which might even be 0!!) 

So we can use the class std attribute that already exists as it was set in the prepare data method.

popde/density_estimate.py
```diff
@@ -120,6 +120,73 @@ def prepare_data(self):
...
+        if self.stdize:
+            std_transf = ['stdize' for dim in self.ndim]
+            std_values_for_points = np.std(transf_points, axis=0)  # record the stds
+            std_points = transf.transform_data(transf_points, std_transf)
+        else:
+            std_values_for_points = None
+            std_points = transf_points
+        #rescale
```
tdent, line 152:
```suggestion

        # Rescale each parameter
```

popde/density_estimate.py
```diff
@@ -120,6 +120,73 @@ def prepare_data(self):
...
+            elif option == 'exp':
+                input_Jacobian = np.exp(points[:, i])
+            elif option in ('none', 'None'):
+                input_Jacobian  = 1.0 
+            else:
+                raise ValueError(f"Invalid transformation option at index {i}: {option}")
+
+            #stdardize 
+            if self.stdize is not None:
+                Jacobian_std_factor =  transf.apply_transf(input_Jacobian,  'stdize')
```
tdent, line 174: What does this do?

tdent: I mean mathematically what formula does it implement

JAMSADIQ: Its dividing the Jacobian with its standard deviation.

The code is using for loops to  compute Jacobian in each dimension so that is why I am calling this method from transf module.

tdent: the question is not about the for loops .. the problem is that 'input_Jacobian' is just a float, so it does not even have a std.  We discussed that this should be the std of the training points instead.

popde/density_estimate.py
```diff
@@ -108,18 +108,83 @@ def prepare_data(self):
             self.transf_data = self.data
 
         if self.stdize:
-            std_transf = ['stdize' for dim in self.ndim]
+            std_transf = ['stdize'] *self.ndim
```
tdent, line 111:
```suggestion
            std_transf = ['stdize'] * self.ndim
```
Spaces around operators

popde/density_estimate.py
```diff
@@ -108,18 +108,83 @@ def prepare_data(self):
...
+        else:
+            std_points = transf_points
+
+        # Rescale each parameter
+        if self.rescale is not None:
+            transf_data = transf.transform_data(std_points, self.rescale)       
+        else:
+            transf_data = std_points
+
+        #Evaluate kde on transform points
```
tdent, line 156:
```suggestion
        # Evaluate kde on transform points
```

popde/density_estimate.py
```diff
@@ -108,18 +108,83 @@ def prepare_data(self):
...
+            if option in['log', 'ln']:
+                input_Jacobian = 1.0 / points[:, i]
+            elif option == 'exp':
+                input_Jacobian = np.exp(points[:, i])
+            elif option in ('none', 'None'):
+                input_Jacobian  = 1.0 
+            else:
+                raise ValueError(f"Invalid transformation option at index {i}: {option}")
+
+            #standardize using stds of training data
```
tdent, line 170:
```suggestion
            # Jacobian for training data standardization
```

popde/density_estimate.py
```diff
@@ -108,18 +108,83 @@ def prepare_data(self):
...
+            elif option in ('none', 'None'):
+                input_Jacobian  = 1.0 
+            else:
+                raise ValueError(f"Invalid transformation option at index {i}: {option}")
+
+            #standardize using stds of training data 
+            if self.stdize:
+                Jacobian_std_factor = self.stds[i]
+            else:
+                Jacobian_std_factor = 1.0
```
tdent, line 174:
```suggestion
                std_Jacobian = self.stds[i]
            else:
                std_Jacobian = 1.0
```
Prefer variables to start with lower case like input_Jacobian

popde/density_estimate.py
```diff
@@ -108,18 +108,83 @@ def prepare_data(self):
...
+
+            #rescale based on self.rescale list in each dimension
```
tdent, line 176:
```suggestion
            # Jacobian for rescaling factor
```

popde/density_estimate.py
```diff
@@ -108,18 +108,83 @@ def prepare_data(self):
...
+            if self.rescale is not None:
+                Jacobian_rescale_factor = self.rescale[i]
+            else:
+                Jacobian_rescale_factor = 1.0
+
+            Total_Jacobian =  input_Jacobian * Jacobian_std_factor * Jacobian_rescale_factor
+            kde_vals   *= Total_Jacobian
```
tdent, line 183:
```suggestion
                rescale_Jacobian = self.rescale[i]
            else:
                rescale_Jacobian = 1.0

            kde_vals *= input_Jacobian * std_Jacobian * rescale_Jacobian
```
I don't think it is necessary to define total_Jacobian here in fact

popde/density_estimate.py
```diff
@@ -108,18 +108,83 @@ def prepare_data(self):
...
+        kde_vals = self.evaluate(transf_data)
+
+        # Jacobian of transform
+        for i, option in enumerate(self.input_transf):
+            if option in['log', 'ln']:
+                input_Jacobian = 1.0 / points[:, i]
+            elif option == 'exp':
+                input_Jacobian = np.exp(points[:, i])
+            elif option in ('none', 'None'):
+                input_Jacobian  = 1.0
```
tdent, line 166:
```suggestion
                input_Jacobian = 1.0 
```

tdent:
The calculation looks OK, several suggestions on comments & formatting

tdent:
(btw I think you should be able to just click on 'accept suggestion' at the web interface for the comments/formatting)

popde/density_estimate.py
tdent: Since this is a bit more complicated due to using self.stds in the standardization step, postpone this for later

tdent (approved):

PR #23, opened 2024-07-04
Add class for general multi-dim KDE with rescaling

tdent:
Implement a general KDE with independent bandwidths in different dimensions by rescaling the training point parameters

### DIFF
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -328,11 +328,11 @@ class VariableBwKDEPy(SimpleKernelDensityEstimation):
 
     Methods:
     --------
-    check_dimensionality():
-        check if the data matches the dimensions of KDE.
+    fit_KDEpy():
+        Set up KDE with a general per-point bandwidth using KDEpy.
 
-    evaluate(points):
-        Evaluate the KDE at given data points.
+    evaluate_KDEpy(points):
+        Evaluate the KDE.
     """
     def __init__(self, data, input_transf=None, stdize=False, rescale=None,
                  backend='KDEpy', bandwidth=1., dim_names=None):
@@ -351,3 +351,17 @@ class VariableBwKDEPy(SimpleKernelDensityEstimation):
 
         return density_values
 
+
+class MultiDimRescalingBwKDEPy(VariableBwKDEPy):
+    """
+    Fit and evaluate multi-dimensional Kernel Density Estimation (KDE)
+    using KDEpy, allowing for variable per-point bandwidth and independent
+    rescaling of each dimension
+
+    Methods:
+    --------
+    TBA
+    """
+    % WIP
+    pass
+    
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -120,7 +120,6 @@ class SimpleKernelDensityEstimation:
         else:
             self.kde_data = self.std_data
 
-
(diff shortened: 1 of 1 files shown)


JAMSADIQ (commented):
looks fine
What methods will be  added, other than the methods in VariableBwKDEPy?

popde/density_estimate.py
```diff
@@ -362,6 +360,30 @@ class MultiDimRescalingBwKDEPy(VariableBwKDEPy):
...
+        A general diagonal covariance Gaussian kernel can be written 
+          K(x - X) = const. * (h_1 h_2 ...)^(-1) exp(-(x - X)^T . diag(h_1^2, h_2^2, ...)^(-1) . (x - X) / 2)
+        This may be implemented by changing variables to w_i = x_i / h_i and 
+        using a unit kernel matrix :
+          K(w - W) = const. * exp(-(w - W)^T . diag(1, 1, ...) . (w - W) / 2).
+        More generally, the kernel in 'w' units should be proportional to the
+        unit matrix (e.g. it may vary between data points).
+        """
+        # Check compatibility of input options
+        if stdize:
```
JAMSADIQ, line 375: why standardization i not allowed?

popde/density_estimate.py
```diff
@@ -362,6 +360,30 @@ class MultiDimRescalingBwKDEPy(VariableBwKDEPy):
     --------
     TBA
     """
-    % WIP
-    pass
-    
+    def __init__(self, data, input_transf=None, stdize=False, rescale=None,
+                 backend='KDEpy', bandwidth=1., dim_names=None,
+                 bandwidth_method='oned_isj')
```
JAMSADIQ, line 361: ``    def __init__(self, data, input_transf=None, stdize=False, rescale=None,
                 backend='KDEpy', bandwidth=1., dim_names=None,
                 bandwidth_method='oned_isj'): 
```

Comment (JAMSADIQ) [reply popde/density_estimate.py:361]: : is missing

File: popde/density_estimate.py [hunk 3]
```diff
@@ -351,3 +349,37 @@ def evaluate_KDEpy(self, points):
...
+        unit matrix (e.g. it may vary between data points).
+        """
+        # Check compatibility of input options
+        if stdize:
+            raise ValueError("Can't standardize variables for this class!")
+        if rescale is not None:
+            raise ValueError("Can't specify rescaling for this class!")
+
+        # Use bandwidth formula to rescale dimensions
+        self.rescale = 1. / self.calc_ndim_bandwidth(bandwidth_method)
```
JAMSADIQ, line 381: this function require self.ndim which is not defined before this. I got and error   
`AttributeError: 'MultiDimRescalingBwKDEPy' object has no attribute 'ndim'`
when run.
Can we add a short example of how to call this class?

tdent:
I don't think we need other methods besides the ones that are inherited at the moment .. at least I can't think of any

popde/density_estimate.py
tdent: Because it would not do anything more than the rescaling we are implementing with the bw method
ie we would standardize then compute a bw for each dimension and rescale, but then we would get the same outcome as if we computed the bw on the data without standardizing 
(since they are both just multiplying by a constant factor)

PR #27, opened 2024-07-09
fixing data type for ISJ bandwidth

JAMSADIQ:


popde/density_estimate.py
```diff
@@ -230,7 +230,8 @@ def calc_ndim_bandwidth(self, method='oned_isj'):
         if method == 'oned_isj':
             from KDEpy.bw_selection import improved_sheather_jones as isj
             for i, col in enumerate(self.kde_data.T):
-                nd_bws[i] = isj(col)  # Weighting is also possible, not implemented atm
+                #col should be a data of shape ndata, ndim==1
```
tdent, line 233: This comment is hard to understand, 'col' is what it is but we need to give the KDEpy function a specific shape of input
```suggestion
                # KDEpy function requires an array of shape (n_samples, 1)
```

popde/density_estimate.py
```diff
@@ -230,7 +230,8 @@ def calc_ndim_bandwidth(self, method='oned_isj'):
...
+                nd_bws[i] = isj(col.reshape(len(col), 1)) # Weighting is also possible, not implemented atm
```
tdent, line 234:
```suggestion
                nd_bws[i] = isj(col[:, np.newaxis]) # Weighting is also possible, not implemented atm
```
I think newaxis should be simpler

tdent (approved):

PR #28, opened 2024-07-09
Optimized bandwidth and alpha for KDE with cross validation on data

jam-sadiq:
adding a new class for optimized bandwidth and alpha with cross validation based on log-likelihood value on training data

### DIFF
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -188,8 +188,8 @@ class AdaptiveKDEOptimization(AdaptiveBwKDE):
                 fom_list = [fom_grid[(bw, al)] for al in self.alpha_options]
                 ax.plot(self.alpha_options, fom_list, 
                         label='{0:.3f}'.format(float(bw)))
-                ax.plot(optalpha, best_score, 'ko', linewidth=10, 
-                        label=r'$\alpha={0:.3f}, bw={1:.3f}$'.format(optalpha, float(optbw)))
+            ax.plot(optalpha, best_score, 'ko', linewidth=10, 
+                    label=r'$\alpha={0:.3f}, bw={1:.3f}$'.format(optalpha, float(optbw)))
             ax.set_xlabel(r'$\alpha$', fontsize=18)
             ax.set_ylabel(r'$FOM$', fontsize=18)
             # add legends on top of plot in multicolumns
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -3,7 +3,6 @@ import scipy
 from density_estimate import VariableBwKDEPy 
 from scipy.stats import gmean
 import matplotlib.pyplot as plt
-import operator
 
 class AdaptiveBwKDE(VariableBwKDEPy):
     """
@@ -175,7 +174,7 @@ class KDEOptimization(AdaptiveBwKDE):
 
                 FOM[(bandwidth, alpha)] = score
 
-        optval = max(FOM.items(), key=operator.itemgetter(1))[0]
+        optval = max(FOM, key=lambda k: FOM[k])
         optbw, optalpha  = optval[0], optval[1]
         best_score = FOM[(optbw, optalpha)]
         # set self.bandwidth  and self.alpha  as optimized values anf fit KDE
@@ -292,7 +291,7 @@ class AdaptiveKDELeaveOneOutCrossValidation():
         for bw in bwgrid:
             for alp in alphagrid:
                 fom[(gbw, alp)] = self.loocv(bw, alp)
-        optvalues = max(fom.items(), key=operator.itemgetter(1))[0]
+        optvalues = max(fom, key=lambda k: fom[k])
         self.optbw, self.optalpha = optvalues[0], optvalues[1]
         self.fom_val = fom[(self.optbw, self.optalpha)]
         kdeval = train_eval_kde(samples, x_eval, self.optbw, self.optalpha)
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -133,7 +133,7 @@ class KDEOptimization(AdaptiveBwKDE):
 
     def loo_cv_score(self, bandwidth_val, alpha_val):
         loo = LeaveOneOut() #sklearn way
-        weights = self.weights
+        fom = 0.0
         for train_index, test_index in loo.split(self.data):
             train_data, test_data = self.data[train_index], self.data[test_index]
            
@@ -158,7 +158,8 @@ class KDEOptimization(AdaptiveBwKDE):
         return sum(fom)
 
     def optimize_parameters(self, method='loo_cv', fom_plot=False):
-        best_score = float('inf')
+        import operator
+        #best_score = float('inf')
         best_params = {'bandwidth': None, 'alpha': None}
 
         FOM= {}
@@ -170,10 +171,15 @@ class KDEOptimization(AdaptiveBwKDE):
                     score = self.loo_cv_score(bandwidth, alpha)
 
                 FOM[(bandwidth, alpha)] = score
-                if score < best_score:
-                    best_score = score
-                    best_params['bandwidth'] = bandwidth
-                    best_params['alpha'] = alpha
+                #if score < best_score:
+                #    best_score = score
+                #    best_params['bandwidth'] = bandwidth
+                #    best_params['alpha'] = alpha
+
+        optval = max(FOM.items(), key=operator.itemgetter(1))[0]
+        optbw, optalpha  = optval[0], optval[1]
+        maxFOM = FOM[(optbw, optalpha)]
+
         
         if fom_plot==True:
             import operator
@@ -205,9 +211,9 @@ class KDEOptimization(AdaptiveBwKDE):
         #set self bandwidth  and alpha
         self.bandwidth  = optbw
         self.alpha  = optalpha
+        best_params = {'bandwidth': optbw, 'alpha': optalpha}
 
-        return best_params, best_score
-
+        return  best_params, maxFOM #best_params, best_score
 
 
 class AdaptiveKDELeaveOneOutCrossValidation():
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -44,6 +44,7 @@ class AdaptiveBwKDE(VariableBwKDEPy):
         # Set up initial KDE with fixed bandwidth
         super().__init__(data, weights, input_transf, stdize,
                          rescale, backend, bandwidth, dim_names, do_fit)
+        #print("in adaptive KDE:", dim_names)
 
         # Compute pilot kde values at input points
         self.pilot_values = self.evaluate(self.kde_data)
@@ -118,47 +119,35 @@ class AdaptiveBwKDE(VariableBwKDEPy):
 
 
 class KDEOptimization(AdaptiveBwKDE):
-    def __init__(data, weights, bandwidth_options, alpha_options , input_transf=None, stdize=False, rescale=None, backend='KDEpy', bandwidth=0.5, alpha=0.0, dim_names=None, do_fit=True)):
-        
+    def __init__(self, data, weights, bandwidth_options, alpha_options , input_transf=None, stdize=False, rescale=None, backend='KDEpy', bandwidth=0.5, alpha=0.0, dim_names=None, do_fit=True):
+        print("dim_names", dim_names)
         self.alpha_options = alpha_options
         self.bandwidth_options = bandwidth_options
 
         super().__init__(data, weights, input_transf, stdize,
                          rescale, backend, bandwidth, dim_names, do_fit)
+        print("super dim_names", dim_names)
     def loo_cv_score(self, bandwidth_val, alpha_val):
         loo = LeaveOneOut() #sklearn way
         fom = 0.0
         for train_index, test_index in loo.split(self.data):
             train_data, test_data = self.data[train_index], self.data[test_index]
             #kde evaluate here with KDEpy
-            kde = AdaptiveBwKDE(train_data, bandwidth=bandwidth_val,alpha=alpha_val)
+            local_weights = np.ones_like(train_data.shape[0])
+            kde = AdaptiveBwKDE(train_data, local_weights, bandwidth=bandwidth_val,alpha=alpha_val)
             fom += kde.evaluate(test_data)  
-    return np.mean(fom)
+        return np.mean(fom)
 
-    def optimize_parameters(self):
-        best_score = float('inf')
-        best_params = {'bandwidth': None, 'alpha': None}
-
-        for alpha in self.alph:
-            for bandwidth in self.bandwidth_options:
-                score = self.loo_cv_score(bandwidth, alpha)
-
-                if score < best_score:
-                    best_score = score
-                    best_params['bandwidth'] = bandwidth
-                    best_params['alpha'] = alpha
-
-    return best_params, best_score
-
-    def kfold_cv_score(self, bandwidth, alpha):
+    def kfold_cv_score(self, bandwidth_val, alpha_val):
         kf = KFold(n_splits=self.n_splits, shuffle=True, random_state=42)
         fom = []
 
         for train_index, test_index in kf.split(self.data):
             train_data, test_data = self.data[train_index], self.data[test_index]
-            kde = AdaptiveBwKDE(train_data, bandwidth=bandwidth_val,alpha=alpha_val)
+            local_weights = np.ones_like(train_data.shape[0])
+            kde = AdaptiveBwKDE(train_data, local_weights, bandwidth=bandwidth_val,alpha=alpha_val)
             fom.append(kde.evaluate(test_data).sum())
-    return sum(fom)
+        return sum(fom)
 
     def optimize_parameters(self, method='loo_cv', fom_plot=False):
         best_score = float('inf')
@@ -167,8 +156,8 @@ class KDEOptimization(AdaptiveBwKDE):
         FOM= {}
         for bandwidth in self.bandwidth_options:
             for alpha in self.alpha_options:
-                if method='kfold_cv':
-                    score = kfold_cv_score(bandwidth, alpha)
+                if method=='kfold_cv':
+                    score = self.kfold_cv_score(bandwidth, alpha)
                 else:
                     score = self.loo_cv_score(bandwidth, alpha)
 
@@ -195,7 +184,7 @@ class KDEOptimization(AdaptiveBwKDE):
                     ax.plot(self.alpha_options, FOMlist, label='{0:.3f}'.format(bw))
                 else:
                     ax.plot(alphagrid, FOMlist, label='{}'.format(bw))
-            if optbw not in ['silverman', 'scott']::
+            if optbw not in ['silverman', 'scott']:
                 ax.plot(optalpha, maxFOM, 'ko', linewidth=10, label=r'$\alpha={0:.3f}, bw= {1:.3f}$'.format(optalpha, optbw))
             else:
                 ax.plot(optalpha, maxFOM, 'ko', linewidth=10, label=r'$\alpha={0:.3f}, bw= {1}$'.format(optalpha, optbw))
@@ -205,13 +194,13 @@ class KDEOptimization(AdaptiveBwKDE):
             lgd = ax.legend(handles, labels, loc='upper center', bbox_to_anchor=(0.5,1.25), ncol =6, fancybox=True, shadow=True, fontsize=8)
         #plt.ylim(maxFOM -5 , maxFOM +6)
             plt.savefig("FOMfortwoDsourcecase.png", bbox_extra_artists=(lgd, ), bbox_inches='tight')
-            plt.close()
+            plt.show()
 
-    #set self bandwidth  and alpha
-    self.bandwidth  = optbw
-    self.alpha  = optalpha
+        #set self bandwidth  and alpha
+        self.bandwidth  = optbw
+        self.alpha  = optalpha
 
-    return best_params, best_score
+        return best_params, best_score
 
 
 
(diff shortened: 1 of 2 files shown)


popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+            plt.close()
+
+        #set self bandwidth  and alpha
+        self.bandwidth  = optbw
+        self.alpha  = optalpha
+
+        return best_params, best_score
+
+
+
```
tdent, line 212:
```suggestion

```

popde/adaptive_kde.py
```diff
@@ -2,6 +2,8 @@
 import scipy
 from density_estimate import VariableBwKDEPy 
 from scipy.stats import gmean
+from sklearn.model_selection import KFold, LeaveOneOut
```
tdent, line 5: it is probably better to import these only when needed, ie inside the relevant methods, since it would be good to be able to run the code for other classes (ie if using a class without grid optimization) without needing to install and import sklearn

popde/adaptive_kde.py
```diff
@@ -2,6 +2,8 @@
...
+import matplotlib.pyplot as plt
```
tdent, line 6: similarly, only import when needed

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
         self.set_alpha(new_alpha)
 
 
+class KDEOptimization(AdaptiveBwKDE):
+    """
+    Given grid of bandwidth and alpha values 
+    find optimized values using cross 
+    valitation with figure of merit
+    based on loglikelihood criteria
```
tdent, line 126:
```suggestion
    Optimize bandwidth and alpha by grid search using
    cross validation with a log likelihood figure of merit 
```

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
 
 
+class KDEOptimization(AdaptiveBwKDE):
+    """
+    Given grid of bandwidth and alpha values 
+    find optimized values using cross 
+    valitation with figure of merit
+    based on loglikelihood criteria 
+    """
+    def __init__(self, data, weights, bandwidth_options, alpha_options , input_transf=None, stdize=False, rescale=None, backend='KDEpy', bandwidth=0.5, alpha=0.0, dim_names=None, do_fit=True):
```
tdent, line 128: I think weights should be a kwarg with default None. 
Also why is bandwidth being given a default value 0.5?  I guess we don't need any default value as this is ignored but 0.5 seems a weird choice ..

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+        self.alpha_options = alpha_options
+        self.bandwidth_options = bandwidth_options
+        super().__init__(data, weights, input_transf, stdize,
+                         rescale, backend, bandwidth, alpha, dim_names, do_fit)
+
+    def loo_cv_score(self, bandwidth_val, alpha_val):
+        loo = LeaveOneOut() #sklearn way
+        weights = self.weights
```
tdent, line 136: this line doesn't seem to do anything ?

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+        for train_index, test_index in loo.split(self.data):
+            train_data, test_data = self.data[train_index], self.data[test_index]
+           
+            #######There may need some hints how to use it better 
+            ### also if we need to add weight option here
+            local_weights = None
```
tdent, line 142: I don't quite understand this comment .. it seems what you are doing is to optimize without using any weights at present, correct ? 
If so, make a comment that explains this and also a 'FIXME' comment to suggest that we might want to use the KDE data weights in future

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+            awkde = AdaptiveBwKDE(train_data, local_weights, bandwidth=bandwidth_val,alpha=alpha_val)
+            fom += awkde.evaluate(test_data)
```
tdent, line 144: fom was not defined earlier, is it a float or a list or what?  .. also does this actually calculate log likelihood?

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+        fom = []
+        for train_index, test_index in kf.split(self.data):
+            train_data, test_data = self.data[train_index], self.data[test_index]
+            local_weights = None
+            awkde = AdaptiveBwKDE(train_data, local_weights, bandwidth=bandwidth_val,alpha=alpha_val)
+            fom.append(awkde.evaluate(test_data).sum())
+        return sum(fom)
+
+    def optimize_parameters(self, method='loo_cv', fom_plot=False):
+        best_score = float('inf')
```
tdent, line 161: I think numpy inf would be preferable ..

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+        best_params = {'bandwidth': None, 'alpha': None}
+
+        FOM= {}
+        for bandwidth in self.bandwidth_options:
+            for alpha in self.alpha_options:
+                if method=='kfold_cv':
+                    score = self.kfold_cv_score(bandwidth, alpha, n_splits=2)
```
tdent, line 168: need n_splits to be controllable via an init option

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+                    score = self.loo_cv_score(bandwidth, alpha)
+
+                FOM[(bandwidth, alpha)] = score
+                if score < best_score:
+                    best_score = score
+                    best_params['bandwidth'] = bandwidth
+                    best_params['alpha'] = alpha
+        
+        if fom_plot==True:
+            import operator
```
tdent, line 179: since operator is a standard python library, I think it's OK to import it at the start of the module

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+            optbw, optalpha  = optval[0], optval[1]
+            maxFOM = FOM[(optbw, optalpha)]
+
+            fig = plt.figure(figsize=(12,8))
+            ax = fig.add_subplot(111)
+            for bw in self.bandwidth_options:
+                FOMlist = [FOM[(bw, al)] for al in self.alpha_options]
+                if bw not in ['silverman', 'scott']:
+                    bw = float(bw) #bwchoice.astype(np.float) #for list
+                    ax.plot(self.alpha_options, FOMlist, label='{0:.3f}'.format(bw))
```
tdent, line 190:
```suggestion
                    ax.plot(self.alpha_options, FOMlist, label='{0:.3f}'.format(float(bw)))
```

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+                else:
+                    ax.plot(alphagrid, FOMlist, label='{}'.format(bw))
+            if optbw not in ['silverman', 'scott']:
```
tdent, line 193: I don't think you need two separate 'if optbw' blocks (here and line 188), can you combine them?

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+                ax.plot(optalpha, maxFOM, 'ko', linewidth=10, label=r'$\alpha={0:.3f}, bw= {1:.3f}$'.format(optalpha, optbw))
+            else:
+                ax.plot(optalpha, maxFOM, 'ko', linewidth=10, label=r'$\alpha={0:.3f}, bw= {1}$'.format(optalpha, optbw))
+            ax.set_xlabel(r'$\alpha$', fontsize=18)
+            ax.set_ylabel(r'$FOM$', fontsize=18)
+            handles, labels = ax.get_legend_handles_labels()
+            lgd = ax.legend(handles, labels, loc='upper center', bbox_to_anchor=(0.5,1.25), ncol =6, fancybox=True, shadow=True, fontsize=8)
+        #plt.ylim(maxFOM -5 , maxFOM +6)
+            plt.savefig("FOMfortwoDsourcecase.png", bbox_extra_artists=(lgd, ), bbox_inches='tight')
```
tdent, line 202: can you explain what this `bbox_extra_artists` is? 
Also, it's not a good idea to always save the plot with the same name, see my suggestion to allow different names to be chosen when calling the method.

popde/adaptive_kde.py
JAMSADIQ: removed

popde/adaptive_kde.py
JAMSADIQ: fixed and called in relevant method

popde/adaptive_kde.py
JAMSADIQ: done

popde/adaptive_kde.py
JAMSADIQ: added as kwarg

JAMSADIQ: bw value is 1 by default

popde/adaptive_kde.py
JAMSADIQ: need more code to fix

popde/adaptive_kde.py
JAMSADIQ: fom  initialized as a float and used log likelihood.

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+        """
+        to do: how to choose and use s_split in init
+        """
+        kf = KFold(n_splits=n_splits, shuffle=True, random_state=42)
+        fom = []
+        for train_index, test_index in kf.split(self.data):
+            train_data, test_data = self.data[train_index], self.data[test_index]
+            local_weights = None
+            awkde = AdaptiveBwKDE(train_data, local_weights, bandwidth=bandwidth_val,alpha=alpha_val)
+            fom.append(awkde.evaluate(test_data).sum())
```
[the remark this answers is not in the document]
JAMSADIQ: now it does, need review if this is correct

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+        FOM= {}
+        for bandwidth in self.bandwidth_options:
+            for alpha in self.alpha_options:
+                if method=='kfold_cv':
+                    score = self.kfold_cv_score(bandwidth, alpha, n_splits=2)
+                else:
+                    score = self.loo_cv_score(bandwidth, alpha)
+
+                FOM[(bandwidth, alpha)] = score
+                if score < best_score:
```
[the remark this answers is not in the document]
JAMSADIQ: corrected with modification. review will be good

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: fix, may  need a  review

popde/adaptive_kde.py
JAMSADIQ: done. KDEpy cannot take scott and silvermann option for multidimensional KDE. need some suggestion here

popde/adaptive_kde.py
```diff
@@ -117,6 +118,98 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+        kf = KFold(n_splits=n_splits, shuffle=True, random_state=42)
+        fom = []
+        for train_index, test_index in kf.split(self.data):
+            train_data, test_data = self.data[train_index], self.data[test_index]
+            local_weights = None
+            awkde = AdaptiveBwKDE(train_data, local_weights, bandwidth=bandwidth_val,alpha=alpha_val)
+            fom.append(awkde.evaluate(test_data).sum())
+        return sum(fom)
+
+    def optimize_parameters(self, method='loo_cv', fom_plot=False):
```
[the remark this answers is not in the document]
JAMSADIQ: fix, need review

popde/adaptive_kde.py
JAMSADIQ: just to fix the legends. added a comment, a suggestion may be better

popde/adaptive_kde.py
```diff
@@ -117,6 +117,100 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+    def __init__(self, data, bandwidth_options, alpha_options, weights=None, input_transf=None, stdize=False, rescale=None, backend='KDEpy', bandwidth=1.0, alpha=0.0, dim_names=None, do_fit=True, n_splits=2):
+        self.alpha_options = alpha_options
+        self.bandwidth_options = bandwidth_options
+        self.n_splits = n_splits
+        super().__init__(data, weights, input_transf, stdize,
+                         rescale, backend, bandwidth, alpha, dim_names, do_fit)
+
+    def loo_cv_score(self, bandwidth_val, alpha_val):
+        from sklearn.model_selection import LeaveOneOut
+        loo = LeaveOneOut() #sklearn way
```
tdent, line 134:
```suggestion
        loo = LeaveOneOut()
```

popde/adaptive_kde.py
```diff
@@ -117,6 +117,100 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+        fom = 0.0
+        for train_index, test_index in loo.split(self.data):
+            train_data, test_data = self.data[train_index], self.data[test_index]
+            #for weights we need corressponding weights as well
```
tdent, line 138: I don't understand this comment ...

popde/adaptive_kde.py
```diff
@@ -117,6 +117,100 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+            if self.weights is not None:
+                local_train_weights = weights[train_index]
+            local_weights = None # FIX ME
+            awkde = AdaptiveBwKDE(train_data, local_weights, bandwidth=bandwidth_val,alpha=alpha_val)
```
tdent, line 142: I am not seeing what the line with local_train_weights is for (also in the kfold case). If the code does nothing at the moment then it is best not to write code at all but instead leave comments discussing possible improvements

popde/adaptive_kde.py
```diff
@@ -117,6 +117,100 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+            fom += np.log(awkde.evaluate(test_data))
+        return np.mean(fom)
```
tdent, line 144: do we need mean here?  It looks like fom is just a float.

popde/adaptive_kde.py
```diff
@@ -117,6 +117,100 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+                else:
+                    score = self.loo_cv_score(bandwidth, alpha)
+
+                FOM[(bandwidth, alpha)] = score
+
+        optval = max(FOM.items(), key=operator.itemgetter(1))[0]
+        optbw, optalpha  = optval[0], optval[1]
+        best_score = FOM[(optbw, optalpha)]
+        #set self.bandwidth  and self.alpha  as optimized values
+        self.bandwidth  = optbw
```
tdent, line 182: do we not want to use `set_adaptive_parameter` here (method of AdaptiveKDE that both sets the parameters and refits the KDE automatically ?)

popde/adaptive_kde.py
```diff
@@ -117,6 +117,100 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
         self.set_alpha(new_alpha)
 
 
+class KDEOptimization(AdaptiveBwKDE):
+    """
+    Optimize bandwidth and alpha by grid search using
+    cross validation with a log likelihood figure of merit 
+    """
+    def __init__(self, data, bandwidth_options, alpha_options, weights=None, input_transf=None, stdize=False, rescale=None, backend='KDEpy', bandwidth=1.0, alpha=0.0, dim_names=None, do_fit=True, n_splits=2):
```
tdent, line 125: I don't think we want `do_fit=True` here, since this would set up the KDE with bw=1 and alpha=0.  Ie we want to initialize all the KDE data but not actually train/fit the KDE until the optimization step.

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: preparing kde with weights  need weight at the training points.  maybe important or disregard for optimized parameter, in progress

popde/adaptive_kde.py
JAMSADIQ: corrected

popde/adaptive_kde.py
JAMSADIQ: correct. done

popde/adaptive_kde.py
JAMSADIQ: removed here in super_init_. need a review

popde/adaptive_kde.py
tdent: but are you saying that you are going to include something to do this in this PR, or not? 
I am saying we should not have 'in progress' code that does not do anything in the repo.
so can we please remove the lines about `local_train_weights` and make notes for any relevant development in another issue or another PR?

popde/adaptive_kde.py
tdent: I am not sure what you need here. We do not need to support 'scott' and 'silverman' in general so I don't know what the problem is

popde/adaptive_kde.py
tdent: I meant that we should have do_fit=False ... have fixed it in an edit myself

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: ok

popde/adaptive_kde.py
```diff
@@ -117,6 +116,88 @@ def set_adaptive_parameter(self, new_alpha, new_global_bw):
...
+                         rescale, backend, bandwidth, alpha, dim_names)
+
+    def loo_cv_score(self, bandwidth_val, alpha_val):
+        from sklearn.model_selection import LeaveOneOut
+        loo = LeaveOneOut() 
+        fom = 0.
+        for train_index, test_index in loo.split(self.data):
+            train_data, test_data = self.data[train_index], self.data[test_index]
+            local_weights = None # FIX ME
+            awkde = AdaptiveBwKDE(train_data, local_weights, bandwidth=bandwidth_val,alpha=alpha_val)
```
tdent, line 140: so here, although we have `input_transf=None, stdize=False, rescale=None` as defaults, it may be that the user initializes the class with some input transformations, and then we will want to apply these to the optimization as well (otherwise we will not get suitable bw/alpha values).  So please also pass on the values of `self.input_transf`, `self.stdize` and `self.rescale` here

tdent (changes requested):
This is basically ready, except for the point about data transformations which need to be passed on to the optimization / FOM calculation.

popde/adaptive_kde.py
JAMSADIQ: so I am bit confuse here. so I can add these for the data but when we evaluate kde on testing data (for score) do we need to also use evaluate_with_transf.

tdent: Can you try to work it out for yourself please ...
We have the original data and we have the transformed data used for KDE fitting
So when doing the cross validation there are some choices that make sense and some that do not
Make one choice that makes sense and code it and I will review the code

JAMSADIQ: I did make changes and tested the code.

popde/adaptive_kde.py
tdent, line 1: I don't see why you have added all these lines here, aren't they already done inside the init method of the parent class ie super.__init__ ?

tdent: I mean lines 130 - 136

tdent (approved):
I think all the issues should have been resolved, if you think it looks OK then please merge

JAMSADIQ:
While testing the code on a twoD data with 
import adaptive_kde as adp_kde
kde = adp_kde.AdaptiveKDEOptimization(data, bwgrid, alpgrid)
I get an error:
    _```
In adaptive_kde.py", line 130, in __init__
    super().__init__(data, weights, input_transf, stdize, rescale, backend,
    adaptive_kde.py", line 47, in __init__
    self.pilot_values = self.evaluate(self.kde_data)
    density_estimate.py", line 370, in evaluate_KDEpy
    density_values = self.kernel_estimate.evaluate(points)
    AttributeError: 'NoneType' object has no attribute 'evaluate
```_

**Comment by tdent** [comment]
OK, so we need to allow `AdaptiveBwKDE` to either have do_fit=True or do_fit=False options.  The problem is that the False option is allowed during super().__init__ but then it is assumed that the kde fit exists.  In any case this class should not produce an error if the option is False.

**Comment by JAMSADIQ** [comment]
Now, with minor fix in density estimate module, this seems to be ready to merge

File: popde/density_estimate.py [hunk 25]
```diff
@@ -172,16 +172,18 @@ def evaluate_with_transf(self, points):
...
         # Jacobian of transforms for each dimension
-        for i, option in enumerate(self.input_transf):
-            if option in['log', 'ln']:
-                input_Jacobian = 1. / points[:, i]
-            elif option == 'exp':
-                input_Jacobian = np.exp(points[:, i])
-            elif option in ('none', 'None'):
-                input_Jacobian = 1. 
+        for i in range(self.ndim):
+            if self.input_transf is not None:
```
tdent, line 176: this should be before the 'for' loop, no?

tdent (approved):
Check if the slight rewrite of the stdize / rescaling Jacobians looks OK, if so then you can go ahead to squash and merge

PR #30, opened 2024-07-22
remove extra lines that were not required

JAMSADIQ:
fixing some extra code

### DIFF
--- a/popde/density_estimate.py
+++ b/popde/density_estimate.py
@@ -67,9 +67,6 @@ class SimpleKernelDensityEstimation:
 
         self.data = np.asarray(data)
         self.ndim = self.data.shape[1]
-        self.weights = np.atleast_1d(weights).astype(float)
-        if len(self.weights) != self.data.shape[0]:
-            raise ValueError("Weights must have same length as data points")
         self.input_transf = input_transf
         self.stdize = stdize
         self.rescale = rescale


popde/density_estimate.py
```diff
@@ -67,9 +67,6 @@ def __init__(self, data, weights=None, input_transf=None, stdize=False,
 
         self.data = np.asarray(data)
         self.ndim = self.data.shape[1]
-        self.weights = np.atleast_1d(weights).astype(float)
```
tdent, line 70: why don't we want to keep these lines but only perform them if weights is not None? 

I.e. 
```
if weights is not None:
    self.weights = (etc...)
    if len(self.weights) != self.data.shape[0]"
        (etc)
else:
    self.weights = None
```

JAMSADIQ: Well, if weights are None than these lines will raise error. These checks are there in next few lines of the code so this is redundant

tdent: I don't know what you mean Jam. I said perform the checks etc if weights is not None. If weights is a valid option for this class then we should perform the checks here and remove the ones done in a subclass.

JAMSADIQ: If you check the next few lines of the code there is this check done. https://github.com/tdent/pop-de/blob/main/popde/density_estimate.py#L82

tdent (approved):

tdent:
Oh. I did not see the other lines just afterwards because github does not display those lines when I am looking at the changes in this PR.

PR #31, opened 2024-08-02
Reorganize adaptive KDE

tdent:
Now use explicit self.pilot_kde assignment to keep fixed bw pilot as separate object, also allow for do_fit to be false


JAMSADIQ:
it looks ok.

JAMSADIQ (approved):
this looks fine.

PR #33, opened 2024-11-06
Opt rescale factor

jam-sadiq:
To accomplish goal for Issue 32:
Add a class for optimize scaling factors for each dimension, along with an optional smoothing parameter (alpha), to improve the performance of a kernel density estimator (KDE). Using Nelder-Mead optimization, it minimizes a cross-validation score based on log-likelihood, helping to fine-tune KDE parameters for accurate density estimation.

### DIFF
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -328,8 +328,9 @@ class KDERescaleOptimization(AdaptiveBwKDE):
     def optimize_rescale_parameters(self, initial_rescale_factor, initial_alpha=0.0, method='kfold_cv', bounds=None):
 
         """
-        Fixed bounds on rescale factor otehrwise it is not working
-        or use better initial guesses using MultiRescaling class
+        given initial choice of rescale factors in each dimension
+        and alpha perform optimization based on 
+        cross validations on data
         """
         #make alpha bounds to be fixed in [0, 1]
         if bounds is None:
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -226,11 +226,9 @@ class AdaptiveKDEOptimization(AdaptiveBwKDE):
 
 class KDERescaleOptimization(AdaptiveBwKDE):
     """
-     A class to optimize rescaling factors for each dimension and 
-      [an optional alpha parameter]
-    using Nelder-Mead optimization. 
-    The optimization is conducted by minimizing a cross-validation
-    score, based on log-likelihood as the figure of merit.
+    A class to optimize rescaling factors for each dimension and alpha 
+    using Nelder-Mead optimization of a cross-validation
+    log-likelihood figure of merit.
 
     Attributes:
         alpha (float): Initial alpha parameter for rescaling.
@@ -238,60 +236,43 @@ class KDERescaleOptimization(AdaptiveBwKDE):
         n_splits (int): Number of splits for k-fold cross-validation.
 
     Methods:
-        loo_cv_score: 
+        loo_cv_score
             Computes Leave-One-Out cross-validation score 
             for given rescale factors.
-        kfold_cv_score(rescale_factors_alpha, seed=42): 
+        kfold_cv_score
             Computes k-fold cross-validation score 
             for given rescale factors.
-        optimize_rescale_parameters(initial_rescale_factor, initial_alpha=0.0, method='loo_cv', fom_plot_name=None, bounds=None):
+        optimize_rescale_parameters
             Optimizes rescaling factors and alpha parameter.
     """
     def __init__(self, data, weights=None, input_transf=None,
-                 stdize=False, rescale=None, backend='KDEpy', bandwidth=1.0, alpha=0.0,
+                 stdize=False, rescale=None, backend='KDEpy', bandwidth=1.0, alpha=0.5,
                  dim_names=None, do_fit=False, n_splits=5):
         """
-        Initialize the KDERescaleOptimization instance 
-        with data and optional configuration parameters.
-
-        Args:
-            data (array-like): Data to fit the KDE model.
-            weights (array-like, optional): Weights for each data point.
-            input_transf (function, optional): Input transformation function.
-            stdize (bool, optional): Whether to standardize data.
-            rescale (array-like, optional): Initial rescale factors 
-               for each dimension.
-            backend (str, optional): Backend to use, default is 'KDEpy'.
+        Args inherited from parent class, except for the following:
+            rescale (array-like, optional): Initial rescale factors for each dimension.
             bandwidth (float, optional): Fixed bandwidth value, default is 1.0.
-            alpha (float, optional): Initial alpha value, default is 0.0.
-            dim_names (list, optional): Dimension names.
-            do_fit (bool, optional): Whether to fit the KDE upon initialization.
-            n_splits (int, optional): Number of splits for k-fold 
-               cross-validation.
+            alpha (float, optional): Initial alpha value.
+            n_splits (int, optional): Number of splits for k-fold cross-validation.
         """
-        self.data = data
         self.n_splits = n_splits
-        self.alpha = alpha
-        self.bandwidth = bandwidth
-        #self.input_transf  = input_transf
-        #self.stdize = stdize
         super().__init__(data, weights, input_transf, stdize, rescale, backend, 
-                bandwidth, alpha, dim_names, do_fit)
+                         bandwidth, alpha, dim_names, do_fit)
 
     def loo_cv_score(self, rescale_factors_alpha):
         """
         Compute the Leave-One-Out Cross-Validation (LOO-CV) score
-          based on log-likelihood.
+          based on log likelihood.
 
         Args:
             rescale_factors_alpha (array-like): Array of rescale factors per
-              dimension and alpha value.
+            dimension and alpha value.
 
         Returns:
-            float: Negative sum of log-likelihood scores for LOO-CV.
+            float: Negative sum of log likelihood for test data.
         """
-        rescale_factors = rescale_factors_alpha[:-1]
-        alpha_choice = rescale_factors_alpha[-1]
+        rescale_val = rescale_factors_alpha[:-1]
+        alpha_val = rescale_factors_alpha[-1]
         from sklearn.model_selection import LeaveOneOut
         loo = LeaveOneOut() 
         fom = 0.
@@ -299,84 +280,72 @@ class KDERescaleOptimization(AdaptiveBwKDE):
             train_data, test_data = self.data[train_index], self.data[test_index]
             local_weights = None # FIX ME
             awkde = AdaptiveBwKDE(train_data, local_weights, input_transf=self.input_transf,
-                                  stdize=self.stdize, rescale=rescale_factors,
-                                  bandwidth=self.bandwidth, alpha=alpha_choice, do_fit=True)
-
+                                  stdize=self.stdize, rescale=rescale_val,
+                                  bandwidth=self.bandwidth, alpha=alpha_val)
             fom += np.log(awkde.evaluate_with_transf(test_data))
+
         return -fom
 
     def kfold_cv_score(self, rescale_factors_alpha, seed=42):
         """
         Perform k-fold cross-validation
         """
-        rescale_factors = rescale_factors_alpha[:-1]
-        alpha_choice = rescale_factors_alpha[-1]
+        rescale_val = rescale_factors_alpha[:-1]
+        alpha_val = rescale_factors_alpha[-1]
         from sklearn.model_selection import KFold
         kf = KFold(n_splits=self.n_splits, shuffle=True, random_state=seed)
         fom = []
-        #this may fix
         for train_index, test_index in kf.split(self.data):
             train_data, test_data = self.data[train_index], self.data[test_index]
             local_weights = None # FIX ME
-
             awkde = AdaptiveBwKDE(train_data, local_weights, input_transf=self.input_transf,
-                                  stdize=self.stdize, rescale=rescale_factors,
-                                  bandwidth=self.bandwidth, alpha=alpha_choice, do_fit=False)
-
-            self.kde_data = self.data
-            awkde.fit_adaptive()
-            #fit now on training data
-            print(awkde.rescale, np.max(test_data), awkde.alpha)
+                                  stdize=self.stdize, rescale=rescale_val,
+                                  bandwidth=self.bandwidth, alpha=alpha_val)
             log_kde_eval = np.log(awkde.evaluate_with_transf(test_data))
             fom.append(log_kde_eval.sum())
             
         return -sum(fom)
 
-    def optimize_rescale_parameters(self, initial_rescale_factor, initial_alpha=0.5, cv_method='kfold_cv', opt_method='Nelder-Mead', bounds=None,  opt_disp_options=None):
-
+    def optimize_rescale_parameters(self, init_rescale=None, init_alpha=None,
+                                    cv_method='kfold_cv', opt_method='Nelder-Mead',
+                                    bounds=None, **kwargs):
         """
-        given initial choice of rescale factors in each dimension
-        and alpha perform optimization based on 
-        cross validations on data
+        Given initial choices of rescale factors in each dimension and alpha,
+        perform optimization with a cross-validated log likelihood FOM
         """
-        #make alpha bounds to be fixed in [0, 1]
+        # Default bounds : alpha must be between 0, 1
         if bounds is None:
-            set_bounds = ((0.01, 100),) * len(initial_rescale_factor) + ((0, 1),)
-        else:
-            set_bounds = bounds
-         # Set default options if none provided
-        disp_options = opt_disp_options if opt_disp_options is None else {'disp': True}
-        initial_choices = np.concatenate((initial_rescale_factor, initial_alpha))
-        from scipy.optimize import minimize, differential_evolution
-        best_params = {'rescale_per_dim': None}
-        # Perform Nelder-mead based Optimization
-        if cv_method == 'kfold_cv':
-            result = minimize(
-                    self.kfold_cv_score,        # func to minimize
-                    initial_choices,             # Initial guess for the parameters
-                    # args= ( )  #Additional arguments to pass to the objective function
-                    method=opt_method,      # Optimization method
-                    options=disp_options,     # Display optimization progress
-                     bounds=set_bounds
-            )
-        elif cv_method == 'loo_cv':
-            result = minimize(
-                    self.loo_cv_score,      
-                    initial_choices,     
-                    #args=(),  # Additional arguments to pass to the objective function
-                    method=opt_method,
-                    options=disp_options , 
-                    bounds =set_bounds
-            )
-        else:
-            raise ValueError("Invalid method specified. Expected 'loo_cv' or 'kfold_cv'.")
+            bounds = ((0.01, 100),) * len(initial_rescale_factor) + ((0., 1.),)
+        # If initial guess is not specified, use class instance values
+        if init_rescale is None:
+            init_rescale = self.rescale
+        if init_alpha is None:
+            init_alpha = self.alpha
+        initial_choices = np.concatenate((init_rescale, init_alpha))
 
-        #make self. rescale be the optimized results
+        try:
+            score_fn = {
+                'kfold_cv': self.kfold_cv_score,
+                'loo_cv': self.loo_cv_score
+            }[cv_method]
+        except KeyError:
+            raise ValueError("Invalid cross-validation method, expected 'loo_cv' or 'kfold_cv'.")
+
+        from scipy.optimize import minimize
+        result = minimize(
+            score_fn,
+            np.concatenate((init_rescale, init_alpha)), # initial guesses
+            method=opt_method,
+            bounds=bounds,
+            options=kwargs  # Pass on additional options as a dictionary
+        )
+
+        # Set instance KDE parameters from the optimized results
         self.rescale = result.x[:-1]
-        self.alpha= result.x[-1]
-        # Return the optimal parameters and the result min value
+        self.alpha = result.x[-1]
+        
         return result.x, result.fun
-            
+
 
 class AdaptiveKDELeaveOneOutCrossValidation():
     """
@@ -466,4 +435,3 @@ class AdaptiveKDELeaveOneOutCrossValidation():
         kdeval = train_eval_kde(samples, x_eval, self.optbw, self.optalpha)
 
         return self.fom_val, self.optbw, self.optalpha
-
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -280,21 +280,19 @@ class KDERescaleOptimization(AdaptiveBwKDE):
         super().__init__(data, weights, input_transf, stdize, rescale, backend, 
                 bandwidth, alpha, dim_names, do_fit)
 
-    def loo_cv_score(self, rescale_factors_alpha):
+    def loo_cv_score(self, rescale_factors, alpha_choice):
         """
         Compute the Leave-One-Out Cross-Validation (LOO-CV) score
           based on log-likelihood.
 
         Args:
-            rescale_factors_alpha (array-like): Array of rescale factors per
+            rescale_factors (array-like): Array of rescale factors per
               dimension and alpha value.
+            alpha_choice (float): float in [0, 1]
 
         Returns:
             float: Negative sum of log-likelihood scores for LOO-CV.
         """
-        alpha = rescale_factors_alpha[-1]
-        rescale_factor = rescale_factors_alpha[0:len(rescale_factors_alpha)-1]
-
         from sklearn.model_selection import LeaveOneOut
         loo = LeaveOneOut() 
         fom = 0.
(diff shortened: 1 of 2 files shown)


tdent:
So I suggested to make a test case for this: make a mock 2d data set with a mixture of two gaussian components at different locations along one coordinate (so that the rate of variation of density is different along different dimensions, even after standardization). 
Then does it choose a different rescaling along the two dimensions and does the resulting KDE look reasonable (more accurate than an isotropic kernel that treats all dimensions equally).

popde/density_estimate.py
```diff
@@ -164,16 +164,24 @@ def calc_ndim_bandwidth(self, method='oned_isj'):
         Assumes a direct (plug-in or comparable) method rather than optimization
         """
         nd_bws = np.zeros(self.ndim)
-        
         # 1-d Botev et al. ("Improved Sheather-Jones") algorithm from KDEpy
-        if method == 'oned_isj':
-            from KDEpy.bw_selection import improved_sheather_jones as isj
+        #if method == 'oned_isj':
```
tdent, line 168: don't leave commented out lines lying around ..

popde/density_estimate.py
```diff
@@ -164,16 +164,24 @@ def calc_ndim_bandwidth(self, method='oned_isj'):
         Assumes a direct (plug-in or comparable) method rather than optimization
         """
         nd_bws = np.zeros(self.ndim)
-        
         # 1-d Botev et al. ("Improved Sheather-Jones") algorithm from KDEpy
```
tdent, line 167: move this comment to where you define oned_isj

popde/density_estimate.py
```diff
@@ -164,16 +164,24 @@ def calc_ndim_bandwidth(self, method='oned_isj'):
...
-        if method == 'oned_isj':
-            from KDEpy.bw_selection import improved_sheather_jones as isj
+        #if method == 'oned_isj':
+
+        from KDEpy.bw_selection import improved_sheather_jones as isj, scotts_rule as scott, silvermans_rule as silverman
+         #dict for mapping method names to functions
```
tdent, line 171: comment is not needed, we can see that it's a dict. 
But since 'methods' is only used once, we don't actually need to define it. 
Ie you could just do 
```
{'oned_isj': isj,
 'scott': scott,
 'silverman': silverman}.get(method)
 ```

File: popde/adaptive_kde.py [hunk 4]
```diff
@@ -224,6 +224,143 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+            for given rescale factors.
+        kfold_cv_score(rescale_factors_alpha, seed=42): 
+            Computes k-fold cross-validation score 
+            for given rescale factors.
+        optimize_rescale_parameters(initial_rescale_factor, initial_alpha=0.0, method='loo_cv', fom_plot_name=None, bounds=None):
+            Optimizes rescaling factors and alpha parameter.
+    """
+    def __init__(self, data, rescale_factors_and_alpha_array, weights=None, input_transf=None,
+                 stdize=False, rescale=None, backend='KDEpy', bandwidth=1.0, alpha=0.0,
+                 dim_names=None, do_fit=False, n_splits=2):
```
tdent, line 252: can the default be 5?

popde/adaptive_kde.py
```diff
@@ -224,6 +224,143 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+
+        Args:
+            rescale_factors_alpha (array-like): Array of rescale factors per
+              dimension and alpha value.
+
+        Returns:
+            float: Negative sum of log-likelihood scores for LOO-CV.
+        """
+        alpha = rescale_factors_alpha[-1]
+        rescale_factor = rescale_factors_alpha[0:len(rescale_factors_alpha)-1]
```
tdent, line 296: can you please look up how to index on the last value of an array or the last but 1 value (etc), numpy allows you to do that with a simple syntax

popde/adaptive_kde.py
```diff
@@ -224,6 +224,143 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+        with data and optional configuration parameters.
+
+        Args:
+            data (array-like): Data to fit the KDE model.
+            rescale_factors_and_alpha_array (array-like): Initial rescale factors
+               and alpha parameter array.
+            weights (array-like, optional): Weights for each data point.
+            input_transf (function, optional): Input transformation function.
+            stdize (bool, optional): Whether to standardize data.
+            rescale (array-like, optional): Initial rescale factors
```
tdent, line 264: I'm not seeing why you need both `rescale_factors_and_alpha_array` and `rescale` and `alpha` ...

popde/adaptive_kde.py
tdent: (that is if you really need to keep this rescale-plus-alpha array which seems a bit awkward)

popde/adaptive_kde.py
```diff
@@ -224,6 +224,143 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+                                  stdize=self.stdize, rescale=rescale_factor,
+                                  bandwidth=self.bandwidth, alpha=alpha)
+            log_kde_eval = np.log(awkde.evaluate(test_data))
+            fom.append(log_kde_eval.sum())
+        return -sum(fom)
+
+    def optimize_rescale_parameters(self, initial_rescale_factor, initial_alpha=0.0, method='loo_cv', fom_plot_name=None, bounds=None):
+
+        """
+        Fixed bounds on rescale factor otehrwise it is not working
```
tdent, line 332: I don't understand this comment. If you need to discuss whether the method works or gives good results please do so on slack or in the comments to this PR, not in code comments.

popde/adaptive_kde.py
```diff
@@ -224,6 +224,143 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+            result = minimize(
+                    self.loo_cv_score,        # why negative
+                    initial_choices,     
+                    #args=(),  # Additional arguments to pass to the objective function
+                    method='Nelder-Mead',
+                    options={'disp': True} #, bounds #crucial maybe
+            )
+
+        #make self. rescale be the optimized results
+        self.rescale = result.x
```
tdent, line 359: I think you need to set alpha as well no?  If alpha is one of the parameters in the optimization you will have Ndim + 1 outputs for 'result.x'

popde/adaptive_kde.py
```diff
@@ -224,6 +224,143 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
         return best_params, best_score
 
 
+class KDERescaleOptimization(AdaptiveBwKDE):
```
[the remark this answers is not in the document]
tdent: hmm, maybe that won't work with how the class is currently, since it is set up to calculate the per-dimension bandwidths with a simple plug-in function rather than take them as input and optimize. OK, forget about this

popde/density_estimate.py

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: removed rescale_factors_and_alpha_array from init

popde/adaptive_kde.py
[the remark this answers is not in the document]
JAMSADIQ: Maybe this idea work. 

Here I am attaching two plots, One where I use this current version of optimizing class on a two d data:  the resulting optimized values of  re-scaling factors are used in KDE and first plot is showing that.
Second plot is just using  _Figure_1_  that compute rescaling factors and fit kde using that.
Figure_2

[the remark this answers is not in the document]
tdent: Sorry, I am not following what you are proposing here.  Again it seems like you have a test data set which does not require different rescaling factors between different dimensions. 
My recommendation at the moment is _not_ to use MultiDimRescalingBwKDEPy and also, as stated in the issue, to do optimization for the rescaling factors.

popde/adaptive_kde.py
```diff
@@ -224,6 +224,147 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+            stdize (bool, optional): Whether to standardize data.
+            rescale (array-like, optional): Initial rescale factors 
+               for each dimension.
+            backend (str, optional): Backend to use, default is 'KDEpy'.
+            bandwidth (float, optional): Fixed bandwidth value, default is 1.0.
+            alpha (float, optional): Initial alpha value, default is 0.0.
+            dim_names (list, optional): Dimension names.
+            do_fit (bool, optional): Whether to fit the KDE upon initialization.
+            n_splits (int, optional): Number of splits for k-fold 
+               cross-validation, default is 2.
```
tdent, line 270: just don't mention the default values and the code will be easier to maintain (here you wrote 2 and then it became wrong)
```suggestion
               cross-validation.
```

popde/density_estimate.py
```diff
@@ -395,8 +398,8 @@ def __init__(self, data, weights=None, input_transf=None, stdize=False,
         unit matrix (e.g. it may vary between data points).
         """
         # Check compatibility of input options
-        if stdize:
-            raise ValueError("Can't standardize variables for this class!")
+        #if stdize:
```
tdent, line 401: don't comment out code that we want to remove, just remove it.

popde/adaptive_kde.py
```diff
@@ -224,6 +224,147 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+    The optimization is conducted by minimizing a cross-validation
+    score, based on log-likelihood as the figure of merit.
+
+    Attributes:
+        alpha (float): Initial alpha parameter for rescaling.
+        bandwidth (float): Fixed bandwidth, default set to 1.0.
+        n_splits (int): Number of splits for k-fold cross-validation.
+
+    Methods:
+        loo_cv_score(rescale_factors_alpha):
```
tdent, line 241: it's not necessary to include the arguments to the method, just leave them out
```suggestion
        loo_cv_score
```

popde/adaptive_kde.py
```diff
@@ -224,6 +224,147 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+        if method == 'kfold_cv':
+            result = minimize(
+                    self.kfold_cv_score,        # func to minimize why negative?
+                    initial_choices,             # Initial guess for the parameters
+                    # args= ( )  #Additional arguments to pass to the objective function
+                    method='Nelder-Mead',      # Optimization method
+                    options={'disp': True}     # Display optimization progress
+                    , bounds=set_bounds
+            )
+        else:
```
tdent, line 350: better to use `elif method == 'loo_cv'` here and then have a value error if it's not either kfold or loo.

JAMSADIQ:
This is a data for testing this issue
Figure_1

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: fixed

popde/adaptive_kde.py
```diff
@@ -224,6 +224,149 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+            set_bounds = ((None, None),) * len(initial_rescale_factor) + ((0, 1),)
+        else:
+            set_bounds = bounds
+        initial_choices = np.concatenate((initial_rescale_factor, initial_alpha))
+        from scipy.optimize import minimize
+        best_params = {'rescale_per_dim': None}
+        # Perform Nelder-mead based Optimization
+        if method == 'kfold_cv':
+            result = minimize(
+                    self.kfold_cv_score,        # func to minimize why negative?
```
tdent, line 343: I don't understand the 'why negative?' comment. Either make it understandable or remove please

tdent:
OK, so can you give the result of calling the new method on the test data, including the call (with initial values / bounds if appropriate), the optimal rescaling/alpha and a plot of the KDE for these optimal parameters

popde/adaptive_kde.py

popde/adaptive_kde.py
```diff
@@ -224,6 +224,149 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+                    initial_choices,             # Initial guess for the parameters
+                    # args= ( )  #Additional arguments to pass to the objective function
+                    method='Nelder-Mead',      # Optimization method
+                    options={'disp': True}     # Display optimization progress
+                    , bounds=set_bounds
+            )
+        elif method == 'loo_cv':
+            print("using leave one out cross validation")
+            result = minimize(
+                    self.loo_cv_score,        # why negative
```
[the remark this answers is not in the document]
JAMSADIQ: ok

JAMSADIQ:
Using the mock data, with initial_guess re-scale factors = [1, 1], alpha =0 , with bounds in optimization function  ((None, None), (None, None),  (0, 1)), the resulting optimize rescale factors = [0.89528626, 0.94384901] and alpha = 1.0 and KDE with these parameters (gloabl_bw=1.0) is shown in the plot
Figure_1
if I change the bounds on re-scaling factors  say min bound is 1 than it shows optimized rescaling factors are interesting and results show distinct features.

JAMSADIQ:
In other experiments  optimized results for re-scaling factors  favors lower bounds, for example if bounds= ((2, 5),(1, 4), (0,1))   the optimized results for re-scale factors =  [2 ,  1.001].

tdent:
So the initial result with the rescale factors close to 1 seems very bad. I don't know why changing the bounds would necessarily help though, except if the bounds happen to push the result closer to the real optimum.
I think the initial values should be more important.
Can you try running with initial values 3, 3, 0 and 0.3, 0.3, 0 to see what happens and keep the bounds wide, like 0.1, 10 in each dimension

Also can you give the cv likelihood value at the end of the minimization

JAMSADIQ:
So, I as I was using minima in optimization I used figure of merit to be negative if log-likelihood.  score in plot title is cross validation value
So this is the test with  initial values 3, 3, 0 and the bounds  0.1, 10 in each dimension. Results are almost same even initial factors are 0.3 for each dimension.
Figure_1

JAMSADIQ:
if I change the figure of merit  with sign and choose initital guess =(0.8, 0.8, 0) and bounds =(0.3, 5) in each dimension that seems to give the plot with score. 
Figure_1

tdent:
I don't see what could be the problem with minimizing -1 * log likelihood such that you might possibly thing that it's a good idea to minimize log likelihood instead. The last result you showed is hitting the rescale factor bounds and it cannot be a good fit to the data either. 
Can you see what happens with initial values [3, 1]?  Using the correct sign of likelihood obviously.

JAMSADIQ:
with bounds (**(3, 10), (1, 10)**, (0,1)) and using initial re-scaling factors [3,1],  with correct log-likelihood sign it seems to choose the lower bound values so there is something problematic in cross validations iterations where kde is estimated.  
Figure_1

tdent:
I cannot see why you are using such tight bounds Jam. The success of the minimization should not depend on the exact tuning of bounds.
However I just had a thought that we are evaluating the likelihood in a wrong way that depends on the rescaling, like with smaller rescaling we get higher densities in the new coordinates and so higher likelihood ..
First have you checked that you are using the right method to evaluate the KDEs at either the original points or the rescaled ones?

tdent:
At the moment you are using the transformed data points (`train_data, test_data = self.kde_data[train_index], self.kde_data[test_index]`) and calculating the KDE and log likelihood in transformed data coordinates (`awkde.evaluate`).

This introduces a difference in the log likelihood FOM compared to calculating in the original or standardized coordinates. If the rescaling was constant then the difference would be constant and have no effect on the optimization, but here that is not the case because we allow the rescaling to vary.  So, here we do not want the log likelihood to change just by rescaling the coordinates (without otherwise changing the KDE). 

So I would recommend doing the FOM evaluation differently by using the original data points and getting the KDE in the original coordinates which are not affected by rescaling (ie `awkde.evaluate_with_transf` using `self.data`).

JAMSADIQ:
ok

JAMSADIQ:
a script to test this class.
```
import numpy as np
import sys
import density_estimate as d
import adaptive_kde as ad
import matplotlib.pyplot as plt
from matplotlib.colors import LogNorm

###Mock data muligaussian peak
np.random.seed(0)
x_centers = [20, 80]  # x dim
x_width = 8           #std 
y_centers = [45, 55]  # y dim
y_width = 20          # ~3*x
num_points = 25

# Generate data for the two clusters
data_x1 = np.random.normal(loc=x_centers[0], scale=x_width, size=num_points)
data_y1 = np.random.normal(loc=y_centers[0], scale=y_width, size=num_points)
data_x2 = np.random.normal(loc=x_centers[1], scale=x_width, size=num_points)
data_y2 = np.random.normal(loc=y_centers[1], scale=y_width, size=num_points)

# to stay within 3,100 range
m1_data = np.clip(np.concatenate([data_x1, data_x2]), 3, 100)
m2_data = np.clip(np.concatenate([data_y1, data_y2]), 3, 100)
plt.figure()
sample = np.column_stack((m1_data, m2_data))
plt.scatter(m1_data, m2_data, alpha=0.8, edgecolors='k')
plt.xlabel("x1")
plt.ylabel("x2")
plt.xlim(0, 100)
plt.ylim(0, 100)
plt.title("mockdata")
plt.show()
#########################################################
#For KDE
sample = np.column_stack((m1_data, m2_data))

m1grid = np.linspace(0, 100 , 200)
m2grid = np.linspace(0, 100, 200)
MM1, MM2 = np.meshgrid(m1grid, m2grid)
grid_pts =  np.array(list(map(np.ravel, [MM1, MM2]))).T
parameter = ['m1', 'm2']
### Initial guess
alphachoice = [0.]
rescale_arr = [0.3, 0.3]
kde2 = ad.KDERescaleOptimization(sample, n_splits=10, stdize=True, input_transf=['none', 'none'], dim_names=['m1source', 'm2source'])
boundvals = ((0.1, 15), (0.1,15), (0,1))
dictopt, score = kde2.optimize_rescale_parameters(rescale_arr, alphachoice, bounds=boundvals)
print(dictopt, score)
print("evaluate KDE based on rescaling factors")
kdenew  = ad.AdaptiveBwKDE(sample, weights=None, bandwidth=1.0, alpha=dictopt[-1], input_transf=['none', 'none'], stdize=True, rescale=dictopt[0:-1])

ZZ = kdenew.evaluate_with_transf(grid_pts).reshape(MM.shape)
print(ZZ.shape)
fig, ax = plt.subplots(figsize=(8, 5), sharey=True)
ax.plot(m1_data, m2_data, 'r+', alpha=0.8)
p = ax.pcolormesh(MM1, MM2 , ZZ, cmap=plt.cm.get_cmap('Purples'))#, norm=LogNorm())
CS = ax.contour(MM1, MM2 , ZZ, colors='black', levels=10, linestyles='dashed', linewidths=1)#, norm=LogNorm())#,levels = contourlevels)
cbar = plt.colorbar(p, ax=ax)
cbar.ax.tick_params(labelsize=20)
ax.tick_params(labelsize=15)
ax.set_xlabel(r'$x_1$', fontsize=16)
ax.set_ylabel(r"$x_2$", fontsize=16)
ax.set_title(f'rescale-factors={dictopt[0:-1]}, score {score}')
fig.tight_layout()
plt.show()
```

popde/density_estimate.py
```diff
@@ -232,6 +243,7 @@ def evaluate_with_transf(self, points):
             std_points = transf_points
 
         # Rescaling
+        #Transform data x' = x / c to shrinks or stretches the data
```
tdent, line 246: it's not clear whether the rescaling factor (values of self.rescale) are notated as 'c' here or as '1/c'. Clarify please and check in transform_utils.py.

popde/density_estimate.py
```diff
@@ -256,8 +268,8 @@ def evaluate_with_transf(self, points):
         # Jacobian for training data standardization
         std_Jacobian = 1. / np.prod(self.stds) if self.stdize else 1.
 
-        # Jacobian for rescaling factor
-        rescale_Jacobian = 1. / np.prod(self.rescale) if (self.rescale is not None) else 1.
+        # Jacobian for rescaling factor f(x) = f(x'*c) = f(x')*c for each dim
```
tdent, line 271: The derivation here seems wrong. 
If F(x') is the KDE for rescaled variables and if x = x' * c, then f(x) is F(x') * c^-1. 

Recall that $\int f(x) dx$ has to be equal to $\int F(x') dx'$.  If the variables get larger the KDE has to get smaller.

popde/density_estimate.py
JAMSADIQ: fixed.

popde/density_estimate.py
JAMSADIQ: fixed now

popde/adaptive_kde.py
```diff
@@ -224,6 +224,151 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+
+            awkde = AdaptiveBwKDE(train_data, local_weights, input_transf=self.input_transf,
+                                  stdize=self.stdize, rescale=rescale_factors,
+                                  bandwidth=self.bandwidth, alpha=alpha_choice, do_fit=True)
+
+            log_kde_eval = np.log(awkde.evaluate_with_transf(test_data))
+            fom.append(log_kde_eval.sum())
+        return -sum(fom)
+
+    def optimize_rescale_parameters(self, initial_rescale_factor, initial_alpha=0.0, method='kfold_cv', bounds=None):
```
tdent, line 327: I think we prefer initial alpha to be nonzero since it may not be good for the optimization if we initialize with a value at the exact boundary.

popde/adaptive_kde.py
```diff
@@ -224,6 +224,151 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+
+        """
+        given initial choice of rescale factors in each dimension
+        and alpha perform optimization based on 
+        cross validations on data
+        """
+        #make alpha bounds to be fixed in [0, 1]
+        if bounds is None:
```
tdent, line 335: I think it's OK to supply some default bounds like (0.01, 100) in each dimension as it's unlikely that the rescaling would go outside these.

popde/adaptive_kde.py
```diff
@@ -224,6 +224,151 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+            result = minimize(
+                    self.kfold_cv_score,        # func to minimize
+                    initial_choices,             # Initial guess for the parameters
+                    # args= ( )  #Additional arguments to pass to the objective function
+                    method='Nelder-Mead',      # Optimization method
+                    options={'disp': True}     # Display optimization progress
+                    , bounds=set_bounds
+            )
+        elif method == 'loo_cv':
+            print("using leave one out cross validation")
```
tdent, line 353: I don't think we want this print statement in production code .. 
```suggestion
```

popde/adaptive_kde.py
tdent: Also, in this method we should have `**kwargs` for additional options that we pass on to the minimize function (like tolerance, printing progress etc.)

popde/adaptive_kde.py
```diff
@@ -224,6 +224,151 @@ def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
...
+        from scipy.optimize import minimize
+        best_params = {'rescale_per_dim': None}
+        # Perform Nelder-mead based Optimization
+        if method == 'kfold_cv':
+            result = minimize(
+                    self.kfold_cv_score,        # func to minimize
+                    initial_choices,             # Initial guess for the parameters
+                    # args= ( )  #Additional arguments to pass to the objective function
+                    method='Nelder-Mead',      # Optimization method
+                    options={'disp': True}     # Display optimization progress
```
tdent, line 349: Rather than hard-code this (ie we probably don't want the code to print the optimization progress hundreds of times if doing iterative KDE), pass on `**kwargs` from the options of the method to make it controllable from a script

popde/adaptive_kde.py
tdent: Also, we might want to control both the cross validation method and the optimization method.  So use eg `cv_method=` and `opt_method=` kwargs

popde/adaptive_kde.py

popde/adaptive_kde.py
JAMSADIQ: done

popde/adaptive_kde.py

popde/adaptive_kde.py

tdent:
So with the new changes it is not necessary to make a new class instance to evaluate with the optimized parameters. Here is what I have for a test script after imports & generating the test data
```
# Initial guesses
alphachoice = 0.5
rescale_arr = [1., 1.]
kde = ad.KDERescaleOptimization(sample, stdize=True, rescale=rescale_arr, alpha=alphachoice, dim_names=['x1', 'x2'])
print('stds', kde.stds)
dictopt, score = kde.optimize_rescale_parameters(rescale_arr, alphachoice, disp=True, xatol=0.01, fatol=0.1)
print(dictopt, score)

print("evaluate KDE based on rescaling factors")
m1grid = np.linspace(0, 100, 100)
m2grid = np.linspace(0, 100, 100)
MM1, MM2 = np.meshgrid(m1grid, m2grid)
eval_pts = np.column_stack([MM1.ravel(), MM2.ravel()])
ZZ = kde.evaluate_with_transf(eval_pts).reshape(MM1.shape)

fig, ax = plt.subplots(figsize=(8, 5), sharey=True)
ax.plot(m1_data, m2_data, 'r+', alpha=0.8)
p = ax.pcolormesh(MM1, MM2, ZZ, cmap=plt.cm.get_cmap('Purples'))
CS = ax.contour(MM1, MM2, ZZ, colors='black', levels=8, linestyles='dashed', linewidths=1)
cbar = plt.colorbar(p, ax=ax)
cbar.ax.tick_params(labelsize=14)
ax.tick_params(labelsize=14)
ax.set_xlabel(r'$x_1$', fontsize=15)
ax.set_ylabel(r"$x_2$", fontsize=15)
ax.set_title(f'rescale factors {dictopt[0:-1]}, alpha {dictopt[-1]:.3f}, lnL {-score:.2f}')
fig.tight_layout()
plt.show()
```

tdent (changes requested):
The new rescaling optimization is good as far as I can see, I will fix the explanation of the rescaling Jacobian as it could still be a bit more straightforward.

tdent:
Please check the new comment text about the stdize and rescaling Jacobians, if it seems correct then go ahead and squash-merge.

tdent (approved):
Please check the new comment text about the stdize and rescaling Jacobians, if it seems correct then go ahead and squash-merge.

PR #34, opened 2025-04-07
 uniform_mass_rescale or bandwidth for  mass dimensions

JAMSADIQ:
adding option for bandwidths in mass dimensions to be same

### DIFF
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -152,7 +152,14 @@ class AdaptiveKDEOptimization(AdaptiveBwKDE):
         super().__init__(data, weights, input_transf, stdize, rescale, backend,
                          bandwidth, alpha, dim_names, do_fit)
 
-    def loo_cv_score(self, bw_val, alpha_val):
+    def loo_cv_score(self, bw_val, alpha_val, uniform_mass_bandwidth=False):
+        # Ensure bw_val is a numpy array
+        bw_val = np.array(bw_val)
+
+        # Apply shared bandwidth for first two dimensions if requested
+        if uniform_mass_bandwidth and len(bw_val) >= 2:
+            bw_val[1] = bw_val[0]  # Set 2nd dim equal to the 1st
+
         from sklearn.model_selection import LeaveOneOut
         loo = LeaveOneOut() 
         fom = 0.
@@ -165,10 +172,16 @@ class AdaptiveKDEOptimization(AdaptiveBwKDE):
             fom += np.log(awkde.evaluate(test_data))
         return fom
 
-    def kfold_cv_score(self, bw_val, alpha_val, seed=42):
+    def kfold_cv_score(self, bw_val, alpha_val, seed=42, uniform_mass_bandwidth=False):
         """
         Perform k-fold cross-validation
         """
+        # Ensure bw_val is a numpy array
+        bw_val = np.array(bw_val)
+
+        # Apply shared bandwidth for first two dimensions if requested
+        if uniform_mass_bandwidth and len(bw_val) >= 2:
+            bw_val[1] = bw_val[0]  # Set 2nd dim equal to the 1st
         from sklearn.model_selection import KFold
         kf = KFold(n_splits=self.n_splits, shuffle=True, random_state=seed)
         fom = []
@@ -182,7 +195,7 @@ class AdaptiveKDEOptimization(AdaptiveBwKDE):
             fom.append(log_kde_eval.sum())
         return sum(fom)
 
-    def optimize_parameters(self, method='loo_cv', fom_plot_name=None):
+    def optimize_parameters(self, method='loo_cv', fom_plot_name=None, uniform_mass_bandwidth=False):
         best_params = {'bandwidth': None, 'alpha': None}
 
         # Perform grid search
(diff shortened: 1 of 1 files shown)


popde/adaptive_kde.py
```diff
@@ -250,15 +251,17 @@ class KDERescaleOptimization(AdaptiveBwKDE):
...
                  stdize=False, rescale=None, backend='KDEpy', bandwidth=1.0, alpha=0.5,
-                 dim_names=None, do_fit=False, n_splits=5):
+                 dim_names=None, do_fit=False, n_splits=5, uniform_mass_rescale=False):
         """
         Args inherited from parent class, except for the following:
             rescale (array-like, optional): Initial rescale factors for each dimension.
             bandwidth (float, optional): Fixed bandwidth value, default is 1.0.
             alpha (float, optional): Initial alpha value.
             n_splits (int, optional): Number of splits for k-fold cross-validation.
+            uniform_mass_rescal(optional, bol): If True first two dimensions will have same bw
```
tdent, line 261: I do not think the option should be called anything specific to this application like 'mass'. The functionality simply sets two hyperparameters to be equal to each other so it should be implemented in a general way without assuming the dimensions are 'mass' or whatever.

tdent: Also, it would be better if the option allows for any set of KDE hyperparameters to be set equal to each other, not just the two first ones in the sequency.

popde/adaptive_kde.py
```diff
@@ -152,7 +152,7 @@ def __init__(self, data, bandwidth_options, alpha_options, weights=None, input_t
         super().__init__(data, weights, input_transf, stdize, rescale, backend,
                          bandwidth, alpha, dim_names, do_fit)
 
-    def loo_cv_score(self, bw_val, alpha_val):
+    def loo_cv_score(self, bw_val, alpha_val, uniform_mass_bandwidth=False):
```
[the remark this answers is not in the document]

popde/adaptive_kde.py
JAMSADIQ: make some changes as suggested

popde/adaptive_kde.py
```diff
@@ -250,15 +251,19 @@ class KDERescaleOptimization(AdaptiveBwKDE):
     """
     def __init__(self, data, weights=None, input_transf=None,
                  stdize=False, rescale=None, backend='KDEpy', bandwidth=1.0, alpha=0.5,
-                 dim_names=None, do_fit=False, n_splits=5):
+                 dim_names=None, do_fit=False, n_splits=5, uniform_rescale=False, uniform_dims=None):
```
tdent, line 254: why do you need two extra arguments here?  I think you could just supply uniform_dims and if this is None do nothing.
Also is there any reason why uniform_dims should have exactly 2 dimensions, could it be used for any number of dimensions?

popde/adaptive_kde.py
```diff
@@ -291,6 +296,21 @@ def loo_cv_score(self, rescale_factors_alpha):
             float: Negative sum of log likelihood over test data.
         """
         rescale_val = rescale_factors_alpha[:-1]
+        # Ensure rescale_val is a numpy array
+        rescale_val = np.array(rescale_val)
+
+        # Apply shared bandwidth 
+        if self.uniform_rescale:
+            assert all(item in self.dim_names for item in self.uniform_dims), "Some elements in cut_seq are not in seq"
```
tdent, line 304: I don't understand this error message, 'cut_seq' and 'seq' are not related to our code

popde/adaptive_kde.py
```diff
@@ -291,6 +296,21 @@ def loo_cv_score(self, rescale_factors_alpha):
             float: Negative sum of log likelihood over test data.
         """
         rescale_val = rescale_factors_alpha[:-1]
+        # Ensure rescale_val is a numpy array
+        rescale_val = np.array(rescale_val)
```
tdent, line 300: can this not be combined with the previous line?  Also the comment is unnecessary, it is obvious that using the `np.array()` function ensures is a numpy array

popde/adaptive_kde.py
```diff
@@ -343,6 +379,16 @@ def optimize_rescale_parameters(self, init_rescale=None, init_alpha=None,
         Given initial choices of rescale factors in each dimension and alpha,
         perform optimization with a cross-validated log likelihood FOM
         """
+        if self.uniform_rescale:
+            assert all(item in self.dim_names for item in self.uniform_dims), "Some elements in cut_seq are not in seq"
```
tdent, line 383: same issue with this opaque error message

popde/adaptive_kde.py
```diff
@@ -355,28 +401,29 @@ def optimize_rescale_parameters(self, init_rescale=None, init_alpha=None,
...
         except KeyError:
             raise ValueError("Invalid cross-validation method, expected 'loo_cv' or 'kfold_cv'.")
-
         from scipy.optimize import minimize
         result = minimize(
             score_fn,
             initial_choices,
             method=opt_method,
-            bounds=bounds,
+            bounds=bounds,
```
tdent, line 416:
```suggestion
            bounds=bounds,
```

tdent (changes requested):
See comments.  Basically this is too specialized and fragile as it assumes that we want exactly two dimensions to have the same bw.  Also you have the same piece of code for removing or adding a parameter several times. 
So I would think of something like this .. we want the class to have an attribute which is a sequence of KDE bws or rescale factors in parallel with the dimension names (possibly including alpha).  This attribute is used for evaluating the KDE. 
But we also want an object which is a sequence of parameters used for the optimization which can have fewer entries. 
Then we also want a method which translates from the opt parameters to the KDE eval parameters to avoid duplicating bits of code between the score/evaluate methods etc.

PR #35, opened 2025-11-12
add symmetrization option to basic KDE

tdent:
Requires separate 'weights' and 'kde_weights' slots to record the original weight assignment vs the (normalized and possibly symmetrized) one used for doing the KDE calculation. 

Currently under testing.


JAMSADIQ (commented):
In density_estimate script     in the class MultiDimRescalingBwKDEPy(VariableBwKDEPy  when calling VariableBwKDEPy 
as super class the order looks flipped for rescale and symmetrize_dims    at line 452-453

PR #36, opened 2025-11-28
added options for symmetrization to take double of per point bandwidth

jam-sadiq:
I added a function in density_estimate script to take into account of symmetry for per-point-bandwidth and also modify adaptive script for two functions where we use perpoint bandwidth, which may not be needed but just to make sure it works.

### DIFF
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -97,8 +97,8 @@ class AdaptiveBwKDE(VariableBwKDEPy):
             The pilot KDE values at the data point positions.
         """
         self.set_bandwidth(
-                self.global_bandwidth / self._local_bandwidth_factor(pilot_values)
-                )
+            self.global_bandwidth / self._local_bandwidth_factor(pilot_values)
+         )
 
     def set_alpha(self, new_alpha):
         """
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -98,7 +98,7 @@ class AdaptiveBwKDE(VariableBwKDEPy):
         """
         self.set_bandwidth(
             self.global_bandwidth / self._local_bandwidth_factor(pilot_values)
-         )
+        )
 
     def set_alpha(self, new_alpha):
         """
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -96,7 +96,9 @@ class AdaptiveBwKDE(VariableBwKDEPy):
         pilot_values : array-like
             The pilot KDE values at the data point positions.
         """
-        self.set_bandwidth(self.global_bandwidth / self._local_bandwidth_factor(pilot_values))
+        self.set_bandwidth(
+                self.global_bandwidth / self._local_bandwidth_factor(pilot_values)
+                )
 
 
     def set_alpha(self, new_alpha):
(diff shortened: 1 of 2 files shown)


popde/adaptive_kde.py
```diff
@@ -295,7 +294,14 @@ def set_rescale(self, new_rescale):
...
         if self.symm_dims is not None:
-            self.symmetrize_data(self.symm_dims)       
+            self.symmetrize_data(self.symm_dims)
+            # If bandwidth is per-point array, double it after symmetrization
+            if isinstance(self.bandwidth, (list, tuple, np.ndarray)):
+                bw_array = np.asarray(self.bandwidth)
+                # Only double if it matches original data length (hasn't been doubled yet)
+                if bw_array.ndim == 1 and len(bw_array) == len(self.data):
+                    self.bandwidth = np.tile(bw_array, 2)
+
```
tdent: in fact for clarity I would suggest changing the call to `self.bandwidth` in the pilot KDE initialization below to `self.global_bandwidth` - it should be the same value, but then it is clearer that we just keep the same fixed global bw and only change the rescale factors.

popde/density_estimate.py
```diff
@@ -89,6 +88,9 @@ def __init__(self, data, weights=None, input_transf=None, stdize=False,
             #if self.weights.max() > 100. or self.weights.max() < 0.01:
             # Commented out as it's not clear if we ever need the 'unnormalized' case.
             self.normalize_weights()
+        else:
+            # Initialize uniform weights if not provided
+            self.kde_weights = None
```
tdent: I don't see why this is needed but I guess it might be useful for `self.kde_weights` to exist, even if we aren't using weights.  But please remove the comment as it does not describe what the code does (it does not provide 'uniform weights').

popde/density_estimate.py
```diff
@@ -1,7 +1,6 @@
 import numpy as np
 from . import transform_utils as transf
 
-
 class SimpleKernelDensityEstimation:
```
tdent:
```suggestion

class SimpleKernelDensityEstimation:
```

popde/density_estimate.py
```diff
@@ -421,6 +424,24 @@ def evaluate_KDEpy(self, points):
...
+        dimensions, and also double the bandwidth array if it's per-point.
+        """
+        # Call parent's symmetrize_data to handle data and weights
+        super().symmetrize_data(dims)
+
+        # Now handle bandwidth if it's an array
+        # Check if bandwidth is array-like (not a scalar)
+        if isinstance(self.bandwidth, (list, tuple, np.ndarray)):
+            bw_array = np.asarray(self.bandwidth)
+            # Only double if it's a 1D array matching original data length
```
tdent: If it's not 1-dimensional then I don't know what the code can do with it. It should always be 1d, so can we just do eg `assert bw_array.ndim == 1` immediately after defining bw_array ?

popde/density_estimate.py
```diff
@@ -421,6 +423,24 @@ def evaluate_KDEpy(self, points):
 
         return density_values
 
+    def symmetrize_data(self, dims):
+        """
+        Override parent method to also handle per-point bandwidth arrays.
```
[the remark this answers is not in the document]
tdent: However, as this is a 'data preparation' method, can it be placed before the fit and evaluate methods?

popde/adaptive_kde.py
```diff
@@ -295,7 +294,14 @@ def set_rescale(self, new_rescale):
         # Re-initialize KDE data
         self.prepare_data()
         if self.symm_dims is not None:
-            self.symmetrize_data(self.symm_dims)       
+            self.symmetrize_data(self.symm_dims)
+            # If bandwidth is per-point array, double it after symmetrization
```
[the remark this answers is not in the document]
jam-sadiq: ok remove this additional part

popde/adaptive_kde.py
jam-sadiq: in removed these lines

popde/density_estimate.py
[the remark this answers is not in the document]
jam-sadiq: fixed

tdent (approved):
have made a couple of minor edits, otherwise it looks good to merge (squash commits)

PR #37, opened 2025-12-10
Fix order of operations in data preprocessing

tdent:
Need to symmetrize before any other transformations, in particular standardization and scaling


jam-sadiq (approved):
ok

PR #42, opened 2026-01-21
Add support for single-dimension reflection in symmetrize_data

jam-sadiq:
Extends the symmetrize_data method to support reflection symmetry in a single dimension, in addition to the existing two-dimension swap functionality based on issue #40.
### Motivation
The current symmetrize_data method only supports swapping values between two dimensions (e.g., m₁ ↔ m₂ symmetry). However, when masses are parameterized as ln(m_total) - ln(q) or ln(m_chirp) - ln(q), the physical symmetry manifests differently: ln(q) → -ln(q) while leaving m_total or m_chirp unchanged. This is a reflection across zero in a single dimension rather than a swap between two dimensions.

### Changes
Modified symmetrize_data to accept either 1 or 2 dimension indices
Two dimensions (existing behavior): Swaps values between the specified dimensions
One dimension (new behavior): Reflects values across zero for the specified dimension

### DIFF
--- a/popde/adaptive_kde.py
+++ b/popde/adaptive_kde.py
@@ -325,7 +325,6 @@ class KDERescaleOptimization(AdaptiveBwKDE):
         alpha_val = rescale_factors_alpha[-1]
         # For 2-dim symmetrization, make another copy of the value for the first symm dimension
         # and insert it after this dim in the rescale array
-        # For 1-dim symmetrization (reflection), no special rescale treatment needed?
         if self.symm_dims is not None and len(self.symm_dims) == 2:
             val_to_copy = rescale_val[self.symm_dims[0]]
             rescale_val = np.insert(rescale_val, self.symm_dims[1], val_to_copy)
@@ -373,7 +372,6 @@ class KDERescaleOptimization(AdaptiveBwKDE):
         alpha_val = rescale_factors_alpha[-1]
         # For 2-dim symmetrization, make another copy of the value for the first symm dimension
         # and insert it after this dim in the rescale array
-        # For 1-dim symmetrization (reflection), no special rescale treatment needed?
         if self.symm_dims is not None and len(self.symm_dims) == 2:
             val_to_copy = rescale_val[self.symm_dims[0]]
             rescale_val = np.insert(rescale_val, self.symm_dims[1], val_to_copy)
@@ -428,7 +426,6 @@ class KDERescaleOptimization(AdaptiveBwKDE):
 
         # For 2-dim symmetrization, check that the dimensions are treated the same and 
         # then remove one of them from the opt parameters
-        # For 1-dim symmetrization (reflection), no special treatment needed ?
         if self.symm_dims is not None and len(self.symm_dims) == 2:
             firstdim = self.symm_dims[0]; secondim = self.symm_dims[1]
             assert init_rescale[firstdim] == init_rescale[secondim]
@@ -455,7 +452,6 @@ class KDERescaleOptimization(AdaptiveBwKDE):
         # Set instance KDE parameters from the optimized results
         rescales = result.x[:-1]
         # For 2-dim symmetrization: copy the rescale value from first dim to second
-        # For 1-dim symmetrization: no special treatment needed ?
         if self.symm_dims is not None and len(self.symm_dims) == 2:
             val_to_copy = rescales[firstdim]
             rescales = np.insert(rescales, secondim, val_to_copy)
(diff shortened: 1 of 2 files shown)


popde/adaptive_kde.py
```diff
@@ -323,9 +323,10 @@ def loo_cv_score(self, rescale_factors_alpha):
         """
         rescale_val = rescale_factors_alpha[:-1]
         alpha_val = rescale_factors_alpha[-1]
-        # For symmetrization, make another copy of the value for the first symm dimension
+        # For 2-dim symmetrization, make another copy of the value for the first symm dimension
         # and insert it after this dim in the rescale array
-        if self.symm_dims is not None:
+        # For 1-dim symmetrization (reflection), no special rescale treatment needed?
```
tdent: This comment is not necessary, please remove it from all the relevant places.
If you want to ask a question you can do that on the pull request webpage comment thread not inside the code

popde/density_estimate.py
```diff
@@ -131,23 +131,40 @@ def check_dimensionality(self):
...
-            if d == int(dims[0]):
-                 extra_data.append(self.kde_data[:, int(dims[1])])
-            elif d == int(dims[1]):
-                 extra_data.append(self.kde_data[:, int(dims[0])])
-            else:  # Leave other dimensions as they are
-                 extra_data.append(self.kde_data[:, d])
-
-        extra_data = np.vstack(extra_data).T  # Stick arrays together
-        assert(extra_data.shape == self.kde_data.shape), extra_data.shape
+        dims = list(dims) if not isinstance(dims, list) else dims
```
tdent: this seems too complicated. Why is the if/else needed? 
```suggestion
        dims = list(dims)
```

popde/density_estimate.py
```diff
@@ -131,23 +131,40 @@ def check_dimensionality(self):
...
+
+        if len(dims) == 2:
+            # Original behavior: swap two dimensions
```
tdent: It is not useful to someone reading the code to know what is 'original' or 'new', the history has no effect on what the code actually does 
```suggestion
        if len(dims) == 2:  # Swap two dimensions
```

popde/density_estimate.py
```diff
@@ -131,23 +131,40 @@ def check_dimensionality(self):
...
+                    extra_data.append(self.kde_data[:, int(dims[1])])
+                elif d == int(dims[1]):
+                    extra_data.append(self.kde_data[:, int(dims[0])])
+                else:  # Leave other dimensions as they are
+                    extra_data.append(self.kde_data[:, d])
+
+            extra_data = np.vstack(extra_data).T  # Stick arrays together
+
+        elif len(dims) == 1:
+            # New behavior: reflect single dimension across zero
```
tdent:
```suggestion
        elif len(dims) == 1:  # Reflect values across zero
```

popde/density_estimate.py
```diff
@@ -131,23 +131,40 @@ def check_dimensionality(self):
...
+            dim_idx = int(dims[0])
```
tdent: why is `int` needed? If the user gives a value which cannot be used as an index it will be an error anyway. If the value can be used as an index I think `int` is not needed.

tdent (changes requested):
Looks ok in general but comments and some pieces of code can be simplified

popde/adaptive_kde.py
jam-sadiq: removed

popde/density_estimate.py
jam-sadiq: changed

popde/density_estimate.py
jam-sadiq: changed as suggested

popde/density_estimate.py
jam-sadiq: fixed

popde/density_estimate.py
jam-sadiq: removed int and seems to work.

popde/density_estimate.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 from . import transform_utils as transf
+import transform_utils as transf
```
tdent: Why has this been added?

tdent:
OK, I looked at whether floats can be used for numpy array or python list indexing, it seems not. So that is why we already have eg `self.kde_data[:, int(dims[1])]` in the code. So I am going to reinstate the int() for the single dimension case.

tdent:
Can you check that the current code works to change self.kde_data in the way we want it to in the 3 cases of `symmetrize_dims = None`, or `= [0, 1]`, or `= [2]` (for instance).

popde/density_estimate.py
jam-sadiq: removed

jam-sadiq:
Now code works for the the options for symmetrize_dims after fixing a extra bracket issues.

tdent (approved):

PR #43, opened 2026-02-06
Symmetrize after applying initial transformation

tdent:
To apply 1d symmetry to a log transformed variable, make the log transformation first and symmetrize inside prepare_data


jam-sadiq (commented):
tested and works

jam-sadiq (approved):

